PR #1, opened 2019-10-25
Update terraform files

doncem:

### Change description ###

This is the first PR settling the journey towards aat deployment. Jenkinsfile_CNP will be introduced as latest as possible so no deployments will happen. Hence non-breaking changes will be. At all.

What this step introduces?

- managed identity (MI) for each environment currently at present
- key vault (as MI needs to be there)
- `product_group_object_id` is used same as bulk scan one. or should i say `reform_dev_logs` 🤷‍♂. in case this needs to be changed it can be done anytime in the future. preferably before introducing any secrets in the vault

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/aat.tfvars
+++ b/infrastructure/aat.tfvars
@@ -1,0 +1,1 @@
+managed_identity_object_id = "85e00c71-7aae-4c14-86e2-7589f60b3357"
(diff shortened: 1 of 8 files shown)


doncem:
i've enabled travis for pr protection but then remembered i've set up only to run on master. of course - can revisit migration to actions later

PR #2, opened 2019-10-30
Enable Jenkins pipeline

doncem:

### Change description ###

Includes some generic fixes and reviews before applying the `Jenkinsfile_CNP` which enables pipeline builds.

The desired outcome is to get green build at least on PR so feature of protecting PR merges can be enabled

```
[ ] Yes
[x] No
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG APP_INSIGHTS_AGENT_VERSION=2.5.0
 
 FROM busybox as downloader
 
-RUN wget -P /tmp https://github.com/microsoft/ApplicationInsights-Java/releases/download/${APP_INSIGHTS_AGENT_VERSION}/applicationinsights-agent-${APP_INSIGHTS_AGENT_VERSION}.jar
+RUN wget -P /tmp https://github.com/microsoft/ApplicationInsights-Java/releases/download/2.5.0/applicationinsights-agent-2.5.0.jar
 
 # Application image
 
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -15,10 +15,6 @@ withPipeline(type, product, component) {
     builder.gradle('integration')
   }
 
-  after('smoketest:aat') {
-    sh "rm -rf infrastructure/"
-  }
-
   enableDockerBuild()
   enableAksStagingDeployment()
   installCharts()
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -1,0 +1,27 @@
+#!groovy
+
+@Library("Infrastructure")
+import uk.gov.hmcts.contino.GradleBuilder
+
+def type = "java"
+def product = "reform-scan"
+def component = "blob-router"
+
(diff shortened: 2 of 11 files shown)


Jenkinsfile_CNP
```diff
@@ -0,0 +1,27 @@
...
+
+  after('smoketest:aat') {
+    sh "rm -rf infrastructure/"
+  }
+
+  enableDockerBuild()
+  enableAksStagingDeployment()
+  installCharts()
+  enableSlackNotifications(channel)
+  disableLegacyDeployment()
```
nitinprabhu: do we need this ? It was never deployed on ASE ? but never know for DNS we may need it

Jenkinsfile_CNP
```diff
@@ -0,0 +1,27 @@
...
+def channel = '#bsp-build-notices'
+GradleBuilder builder = new GradleBuilder(this, product)
+
+withPipeline(type, product, component) {
+  after('test') {
+    builder.gradle('integration')
+  }
+
+  after('smoketest:aat') {
+    sh "rm -rf infrastructure/"
```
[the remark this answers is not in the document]
doncem: true. left/included this for for super safety i guess :D

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
         - http_proxy
         - https_proxy
         - no_proxy
-    image: hmcts/blob-router-service
+    image: reform-scan/blob-router
```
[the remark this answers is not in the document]
doncem: perhaps needs full url, right?

Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG APP_INSIGHTS_AGENT_VERSION=2.5.0
 
 FROM busybox as downloader
 
-RUN wget -P /tmp https://github.com/microsoft/ApplicationInsights-Java/releases/download/${APP_INSIGHTS_AGENT_VERSION}/applicationinsights-agent-${APP_INSIGHTS_AGENT_VERSION}.jar
+RUN wget -P /tmp https://github.com/microsoft/ApplicationInsights-Java/releases/download/2.5.0/applicationinsights-agent-2.5.0.jar
```
rkondratowicz: please explain

doncem: not sure which part to explain so will try to explain as fully as i can.

commit message says "Fix `wget` command in Dockerfile". diff suggests that previous version did not work. ARG setting is not used in this command line run hence need hardcoded yet again.

it is way easier to change 3 places in this file than downloading jar yourself and sticking in repository.

inspected such error on different application - sample app just moments ago

nitinprabhu: Do we need APP_INSIGHTS_AGENT_VERSION any more than ?

doncem: Can't see in jenkins library. If suppose to be used only by that library then it doesn;t serve any value right now - single point of use below in same file. I can test it out once pipeline is ready in Jenkins. For now - let's keep it as all other projects (including template) has it

PR #3, opened 2019-11-04
Fix pipeline deployment - for PRs and maybe master

doncem:

### Change description ###

Master pipeline is ready. First real bug discovered by it and clearly visible in the initial commit. This PR will try to address any issues along the way to make at least PR pipeline pass

```
[ ] Yes
[x] No
```


doncem:
First PR run on Jenkins - great success. Enabling such requirement in settings then and merging this one so to see next step in master pipeline which actually runs terraform

PR #4, opened 2019-11-05
Add resource for Application Insights and store to vault

doncem:

### Change description ###

Name reference is ~picked up from Jenkins pipeline. It errors to fetch it but is soft dependency so not a failure to deploy.~ standard one

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -43,7 +43,7 @@ resource "azurerm_application_insights" "appinsights" {
 
 # store app insights key in key vault
 resource "azurerm_key_vault_secret" "appinsights_secret" {
-  name         = "AppInsightsInstrumentationKey"
+  name         = "app-insights-instrumentation-key"
   value        = "${azurerm_application_insights.appinsights.instrumentation_key}"
   key_vault_id = "${data.azurerm_key_vault.key_vault.id}"
 }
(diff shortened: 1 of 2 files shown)


infrastructure/main.tf
```diff
@@ -32,3 +32,18 @@ data "azurerm_key_vault" "key_vault" {
...
+resource "azurerm_application_insights" "appinsights" {
+  name                = "${var.product}-${var.env}"
+  location            = "${var.location}"
+  resource_group_name = "${azurerm_resource_group.rg.name}"
+  application_type    = "${var.application_type}"
+  tags                = "${var.common_tags}"
+}
+
+# store app insights key in key vault
+resource "azurerm_key_vault_secret" "bulk_scan_orchestrator_app_s2s_secret" {
```
nitinprabhu: name should be app insights and not secret ?

infrastructure/main.tf
```diff
@@ -32,3 +32,18 @@ data "azurerm_key_vault" "key_vault" {
...
+  name         = "AppInsightsInstrumentationKey"
```
nitinprabhu: can we plz name it `app-insights-instrumentation-key` ?

infrastructure/main.tf
doncem: 🙊 . let's agree together - you didn't see it :D

infrastructure/main.tf
doncem: and how about jenkins pipeline using this? at first i did put in what you suggest

nitinprabhu: that would be set in bootstrap.yaml ?

PR #8, opened 2019-11-08
Provision blob storage

doncem:

[Provision new BLOB](https://tools.hmcts.net/jira/browse/BPS-805)

### Change description ###

Provisioning new blob storage for `reform-scan` product:

- including palo alto
- storing secrets
- creating containers

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/palo-alto.tf
+++ b/infrastructure/palo-alto.tf
@@ -1,0 +1,21 @@
+locals {
+  trusted_vnet_name           = "core-infra-vnet-${var.env}"
+  trusted_vnet_resource_group = "core-infra-${var.env}"
+  trusted_vnet_subnet_name    = "palo-trusted"
+}
+
+module "palo_alto" {
+  source       = "git@github.com:hmcts/cnp-module-palo-alto?ref=master"
+  subscription = "${var.subscription}"
+  env          = "${var.env}"
+  product      = "${var.product}"
+  common_tags  = "${var.common_tags}"
+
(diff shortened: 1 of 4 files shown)


infrastructure/palo-alto.tf
```diff
@@ -0,0 +1,21 @@
+locals {
+  trusted_vnet_name           = "core-infra-vnet-${var.env}"
+  trusted_vnet_resource_group = "core-infra-${var.env}"
+  trusted_vnet_subnet_name    = "palo-trusted"
+}
+
+module "palo_alto" {
```
timja: I would be quite wary of having the palo alto infrastructure in an application repository,
the automation is quite poor, runs on every build and sometimes fails

you probably want the infrastructure provisioned in a separate  repo just for infrastructure?

doncem: ah. so here comes the culprit which requires separate infra. will it be safe to move out all the rest as well? good thing nothing is used atm

nitinprabhu: Agree with @timja else we would getting blocked with every build in future

doncem: ok. will extract what's current and raise this in new repo

timja: it's this guy + the DNS challenges required  for a custom domain on the storage account can be quite a pain...

doncem:
will move to new repo

PR #9, opened 2019-11-08
Remove soon to become shared infra

doncem:
### Change description ###

#8 mentions conflicting module not recommended to have in repository itself since every pipeline run will execute the terraform which is not desired. Will move to new repository but first - need to clear existing setup. No1 is using this service nor this service does any impact - just now got deployed to AAT with "welcome to service" endpoint

```
[x] Yes
[x] No
```


charts/reform-scan-blob-router/values.aat.template.yaml
```diff
@@ -3,4 +3,3 @@ java:
   # Don't modify below here
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
-  devApplicationInsightsInstrumentKey: "4bb6dd82-2362-4899-acba-56e027dddc49"
```
nitinprabhu: this can stay here ?

src/main/resources/bootstrap.yaml
```diff
@@ -5,4 +5,3 @@ spring:
       prefixed: false
       paths: /mnt/secrets/reform-scan
       aliases:
-        app-insights-instrumentation-key: azure.application-insights.instrumentation-key
```
nitinprabhu: this can stay here ?

charts/reform-scan-blob-router/values.aat.template.yaml
doncem: I don't think so. all resources will be deleted. Recreated in another repository so presuming - different IDs. Same answer for next one: since key vault will be deleted - nothing to load from

nitinprabhu: update the new id once you have created new on ? No relation to key vault this is the app insights key 🤔

doncem: oh yes

src/main/resources/bootstrap.yaml
doncem: but this one perhaps i need to `enabled: false`. dunno, pr is green and i would guess it is using bootstrap setup ^

nitinprabhu: you can keep this although app insights key will be still be the old one. Once you have correct value in key vault it will set automatically.

doncem: i mean the vault itself will be deleted so there won't be any value to load from

PR #20, opened 2019-11-15
Add config for blob storage secrets

aliveni:
https://tools.hmcts.net/jira/browse/BPS-869

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -14,8 +14,9 @@ spring:
     name: Blob Router Service
 
 storage:
-  account-name: ${STORAGE_ACCOUNT_NAME:bulkscansandbox}
-  account-key: ${STORAGE_KEY:dGVzdA==} #Base 64 encoded
+  account-name: ${STORAGE_ACCOUNT_NAME:reformscanaat}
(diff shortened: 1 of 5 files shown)


src/main/resources/application.yaml
```diff
@@ -14,8 +14,9 @@ spring:
     name: Blob Router Service
 
 storage:
-  account-name: ${STORAGE_ACCOUNT_NAME:bulkscansandbox}
-  account-key: ${STORAGE_KEY:dGVzdA==} #Base 64 encoded
+  account-name: ${STORAGE_ACCOUNT_NAME:reformscanaat}
```
doncem: I think `sandbox` suffix was fine :)

src/main/resources/bootstrap.yaml
```diff
@@ -6,3 +6,6 @@ spring:
       paths: /mnt/secrets/reform-scan
       aliases:
         app-insights-instrumentation-key: azure.application-insights.instrumentation-key
+        storage-account-name: storage.account_name
+        storage-account-primary-key: storage.account_key
+        storage-account-secondary-key: storage.account_secondary_key
```
doncem: should those be like env vars used in application.yaml?

aliveni: yeah, done

PR #22, opened 2019-11-15
Add smoke test for the Storage sas token

aliveni:
https://tools.hmcts.net/jira/browse/BPS-869

### Change description ###
Add smoke test for SAS token endpoint.

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -217,8 +217,12 @@ dependencies {
   functionalTestCompile group: 'org.assertj', name: 'assertj-core', version: '3.14.0'
 
   smokeTestCompile sourceSets.main.runtimeClasspath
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -217,8 +217,13 @@ dependencies {
   functionalTestCompile group: 'org.assertj', name: 'assertj-core', version: '3.14.0'
 
   smokeTestCompile sourceSets.main.runtimeClasspath
+  smokeTestCompile sourceSets.test.runtimeClasspath
```
doncem: This should not be included. Alter tests to avoid whole lot dependencies from test sources

PR #27, opened 2019-11-25
Add modules for API management

doncem:

[New API in the existing API Management Service](https://tools.hmcts.net/jira/browse/BPS-870)

### Change description ###

Similar (very close) to setup used in bulk scan processor except this one is using new modules provided by platform.

Thumbprints will be introduced together with tests

~NOTE. Making a draft PR as don't know what to do with `revision`. Please advise~

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/api-mgmt.tf
+++ b/infrastructure/api-mgmt.tf
@@ -1,0 +1,48 @@
+locals {
+  # certificate thumbprints used by API tests
+  allowed_test_certificate_thumbprints = [
+    "${var.api_test_valid_certificate_thumbprint}",
+    "${var.api_test_expired_certificate_thumbprint}",
+    "${var.api_test_not_yet_valid_certificate_thumbprint}",
+  ]
+
+  # list of the thumbprints of the SSL certificates that should be accepted by the API (gateway)
+  allowed_certificate_thumbprints = "${concat(
+    local.allowed_test_certificate_thumbprints,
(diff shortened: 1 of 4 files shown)


infrastructure/api-mgmt.tf
```diff
@@ -0,0 +1,48 @@
...
+  api_mgmt_rg   = "core-infra-${var.env}"
+  name          = "${var.component}"
+}
+
+module "api_mgmt" {
+  source        = "git@github.com:hmcts/cnp-module-api-mgmt-api.git"
+  name          = "${var.component}-api"
+  api_mgmt_name = "core-api-mgmt-${var.env}"
+  api_mgmt_rg   = "core-infra-${var.env}"
+  revision      = "1" //???
```
timja: they're all 1 in prod, you can provide different API versions through this I think

infrastructure/api-mgmt.tf
```diff
@@ -0,0 +1,48 @@
...
+  product_id    = "${module.api_mgmt_product.product_id}"
+  display_name  = "Get SAS token for storage account access"
+  path          = "/token/*"
```
timja: not sure with the * is needed but if you got it from somewhere else should be fine

doncem: Yep, same is used in bulk scan processor

PR #30, opened 2019-11-25
Allow disabling containers via config

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-920

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/config/ServiceConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/config/ServiceConfiguration.java
@@ -25,6 +25,7 @@ public class ServiceConfiguration {
     public static class StorageConfig {
         private String name;
         private int sasValidity;
+        private boolean isEnabled = true;
 
         public String getName() {
             return name;
@@ -40,6 +41,14 @@ public class ServiceConfiguration {
 
         public void setSasValidity(int sasValidity) {
             this.sasValidity = sasValidity;
+        }
+
+        public boolean isEnabled() {
+            return isEnabled;
+        }
+
+        public void setEnabled(boolean enabled) {
+            isEnabled = enabled;
         }
     }
 }
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/SasTokenGeneratorService.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/SasTokenGeneratorService.java
@@ -14,6 +14,7 @@ import uk.gov.hmcts.reform.blobrouter.exceptions.ServiceConfigNotFoundException;
 import uk.gov.hmcts.reform.blobrouter.exceptions.UnableToGenerateSasTokenException;
 
 import java.time.OffsetDateTime;
+import java.util.Optional;
 
 @EnableConfigurationProperties(ServiceConfiguration.class)
 @Service
@@ -58,9 +59,9 @@ public class SasTokenGeneratorService {
     }
 
     private StorageConfig getConfigForService(String serviceName) {
-        if (!serviceConfiguration.getStorageConfig().containsKey(serviceName)) {
-            throw new ServiceConfigNotFoundException("No service configuration found for " + serviceName);
-        }
-        return serviceConfiguration.getStorageConfig().get(serviceName);
+        return Optional
+            .ofNullable(serviceConfiguration.getStorageConfig().get(serviceName))
+            .filter(StorageConfig::isEnabled)
+            .orElseThrow(() -> new ServiceConfigNotFoundException("No service configuration found for " + serviceName));
     }
 }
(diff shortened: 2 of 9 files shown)


src/test/java/uk/gov/hmcts/reform/blobrouter/services/SasTokenGeneratorServiceTest.java
```diff
@@ -56,20 +65,26 @@ void should_throw_exception_when_requested_service_is_not_configured() {
             .hasMessage("No service configuration found for nonexistingservice");
     }
 
+    @Test
+    void should_throw_exception_when_service_is_disabled() {
+        assertThatThrownBy(() -> tokenGeneratorService.generateSasToken(DISABLED_SERVICE))
+            .isInstanceOf(ServiceConfigNotFoundException.class)
```
nitinprabhu: should it be disable service or something instead of config not found ?

rkondratowicz: I thought it would be best to not differentiate between those two scenarios.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/SasTokenGeneratorService.java
```diff
@@ -58,9 +59,9 @@ private BlobServiceSasSignatureValues getBlobServiceSasSignatureValues(String se
...
 
     private StorageConfig getConfigForService(String serviceName) {
-        if (!serviceConfiguration.getStorageConfig().containsKey(serviceName)) {
-            throw new ServiceConfigNotFoundException("No service configuration found for " + serviceName);
-        }
-        return serviceConfiguration.getStorageConfig().get(serviceName);
+        return Optional
+            .ofNullable(serviceConfiguration.getStorageConfig().get(serviceName))
+            .filter(StorageConfig::isEnabled)
+            .orElseThrow(() -> new ServiceConfigNotFoundException("No service configuration found for " + serviceName));
```
doncem: Do we want to throw same exception for services which are disabled as well? Exception gets translated into `BAD_REQUEST` btw.

If we do, introducing `Optional` here perhaps is okay, otherwise we could use map functionality alone: `map.computeIfAbsent`

src/test/java/uk/gov/hmcts/reform/blobrouter/services/SasTokenGeneratorServiceTest.java
nitinprabhu: Differentiating  would make it clear in my opinion

doncem: It **is** configured, but it is **disabled**

rkondratowicz: ok, I'll change this

rkondratowicz: done. Added new exception type, used different message

src/main/java/uk/gov/hmcts/reform/blobrouter/config/ServiceConfiguration.java
```diff
@@ -25,6 +25,7 @@ public void setStorageConfig(List<StorageConfig> storageConfig) {
     public static class StorageConfig {
         private String name;
         private int sasValidity;
+        private boolean isEnabled = true;
```
doncem: So shall we name `enabled` instead?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/SasTokenGeneratorService.java
```diff
@@ -58,9 +59,13 @@ private BlobServiceSasSignatureValues getBlobServiceSasSignatureValues(String se
...
     private StorageConfig getConfigForService(String serviceName) {
-        if (!serviceConfiguration.getStorageConfig().containsKey(serviceName)) {
+        StorageConfig config = serviceConfiguration.getStorageConfig().get(serviceName);
+        if (config == null) {
             throw new ServiceConfigNotFoundException("No service configuration found for " + serviceName);
+        } else if (!config.isEnabled()) {
+            throw new ServiceDisabledException("Service " + serviceName + " has been disabled.");
+        } else {
+            return config;
         }
```
doncem: Let us understand this better: previous [commit](https://github.com/hmcts/blob-router-service/pull/30/commits/95a2cf79cf0bddc65e77686dd4c2d94b167b4bfb) it was okay to use `Optional` wrapper together with it's functional capabilities but after introducing another exception we are forcing ourselves to use if/else combination and _hardcoding_ `null` value.

Why not trusting java and use it as library properly plus it will give `+` points from future java releases (and currently used - 11).

```java
StorageConfig config = serviceConfiguration
    .getStorageConfig()
    .computeIfAbsent(serviceName, service -> {
        throw new ServiceConfigNotFoundException("No service configuration found for " + service);
    });

if (config.isDisabled()) {
    throw new ServiceDisabledException("Service " + serviceName + " has been disabled.");
}

return config;
```

Tiny note: at the moment we only need `isDisabled()` function 🙂

rkondratowicz: I much prefer my version

PR #34, opened 2019-12-03
Feature: Application Gateway tests in the pipeline

doncem:

[New API in the existing API Management Service](https://tools.hmcts.net/jira/browse/BPS-870)

### Change description ###

Api is set up (#27 + #28 + #29). Adding basic tests to prove it's working. Extra files had to be changed but overall changes are small and not candidate for separate PR. Especially when need to provide flag switch for enabling app gw tests in pipeline

```
[x] Yes - not sure tests will pass
[x] No - no source code is actually broken/changed
```


infrastructure/api-mgmt.tf
```diff
@@ -30,8 +30,8 @@ module "api_mgmt" {
   api_mgmt_rg   = "core-infra-${var.env}"
   revision      = "1"
   product_id    = "${module.api_mgmt_product.product_id}"
-  display_name  = "Get SAS token for storage account access"
-  path          = "token"
+  display_name  = "Blob Router API"
+  path          = "reform-scan"
```
nitinprabhu: is path change intentional ? Third party supplier will need to update ? This is anyways to retrieve SAS token for should be fine as it is ?

src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter/ApiGatewayTest.java
```diff
@@ -0,0 +1,172 @@
...
+    private static File validJavaKeyStore;
+    private static final String VALID_KEY_STORE_PASSWORD = CONFIG.getString("client.valid-key-store.password");
+
+    @BeforeAll
+    static void checkConfig() throws IOException {
+        assertThat(API_URL).as("API gateway URL").isNotEmpty();
+        assertThat(SUBSCRIPTION_KEY).as("Subscription key").isNotEmpty();
+
+        // create tmp file for valid certificate retrieved from key vault
+        validJavaKeyStore = File.createTempFile("appGW", "test");
```
nitinprabhu: are we sure this file creation will work ?

nitinprabhu: It would be nice if we could similar to https://github.com/hmcts/bulk-scan-processor/blob/master/src/apiGatewayTest/java/uk/gov/hmcts/reform/bulkscanprocessor/api/GetSasTokenTest.java

src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter/ApiGatewayTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        try (FileOutputStream fos = new FileOutputStream(validJavaKeyStore)) {
+            fos.write(Base64.getDecoder().decode(CONFIG.getString("client.valid-key-store.content")));
+        }
+    }
+
+    @Test
+    void should_accept_request_with_valid_certificate_and_subscription_key() {
```
nitinprabhu: can we also please add test case for valid cert and invalid sub key ?

infrastructure/api-mgmt.tf
doncem: yes, intentional. the construct is \<api-gw\>/\<path\>/\<swagger\> so `app-gw-url/reform-scan/token/{service_name}`

src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter/ApiGatewayTest.java
doncem: No it would not be nice as it uses deprecated features. I've spent some time to avoid that and use `SSLConfig` without deprecation. This requires file to be present. We already have non valid one in resources, need to adapt same logic for the one stored in key vault. Tests are pretty much exactly the same as linked above ^ but with slight modification to not use deprecated features

src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter/ApiGatewayTest.java
doncem: That's second (the followin ❗️) test case 😑

infrastructure/api-mgmt.tf
nitinprabhu: Consumer(supplier) would have an impact is it agreed change ?

src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter/ApiGatewayTest.java
nitinprabhu: > are we sure this file creation will work ?

@doncem

src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter/ApiGatewayTest.java
nitinprabhu: :)

infrastructure/api-mgmt.tf
doncem: nobody agreed on anything in the first place. all api gateways require path as it's same url. bulk scan processor has `bulk-scan`, fees - `fees-api` and so on. reform scan product (this component is under that flag) has `reform-scan` after this pr

src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter/ApiGatewayTest.java
doncem: we'll see once merged. but pretty sure it will. it's normal java io library which is valid inside jvm and/or whichever environment it runs (kubes right now)

infrastructure/api-mgmt.tf
nitinprabhu: ok I think there is no choice we will have to update the path

PR #43, opened 2019-12-11
Fixed bootstrap config and removed defaults

nitinprabhu:

https://tools.hmcts.net/jira/browse/BPS-962

### Change description ###

- Removed default values for storage account name and keys and also updated the mappings.

- Currently after merging https://github.com/hmcts/blob-router-service/pull/42  it is still using default values for storage name and key which means bootstrap mappings seems incorrect.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.preview.template.yaml
+++ b/charts/reform-scan-blob-router/values.preview.template.yaml
@@ -14,12 +14,13 @@ java:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
   aadIdentityName:
-  blobstorage:
-    resourceGroup: reform-scan-aks
-    teamName: "BSP"
-    location: ukwest
-    setup:
-      containers:
-        - bulkscan
-        - bulkscan-rejected
-    enabled: true
+
+blobstorage:
+  resourceGroup: reform-scan-aks
+  teamName: "BSP"
+  location: ukwest
+  setup:
+    containers:
+      - bulkscan
+      - bulkscan-rejected
+  enabled: true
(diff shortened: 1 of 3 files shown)


src/main/resources/bootstrap.yaml
```diff
@@ -6,6 +6,6 @@ spring:
       paths: /mnt/secrets/reform-scan
       aliases:
         app-insights-instrumentation-key: azure.application-insights.instrumentation-key
-        storage-account-name: STORAGE_ACCOUNT_NAME
-        storage-account-primary-key: STORAGE_ACCOUNT_KEY
-        storage-account-secondary-key: STORAGE_ACCOUNT_SECONDARY_KEY
+        storage-account-name: storage.account_name
+        storage-account-primary-key:  storage.account_key
+        storage-account-secondary-key:  storage.account_secondary_key
```
doncem: Hmm.. not sure this change is needed 🤔

charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -1,6 +1,26 @@
...
+
+blobstorage:
+  resourceGroup: reform-scan-aks
+  teamName: "BSP"
+  location: ukwest
+  setup:
+    containers:
+      - bulkscan
+      - bulkscan-rejected
+  enabled: true
```
doncem: 👍

src/main/resources/bootstrap.yaml
nitinprabhu: Existing one didn't work so assume this would work. We will know now as the defaults have been removed

PR #46, opened 2019-12-12
Add architecture of blob processing with azure blob storage library v12

doncem:

In between [Create an AKS-triggered job[Spike]](https://tools.hmcts.net/jira/browse/BPS-914) and [Create blob processing component](https://tools.hmcts.net/jira/browse/BPS-919)

### Change description ###

NOTE! Full orchestration so to show the separation which also comes from azure blob storage library.

Just architecture setup as it is quite different to what we are used to with v8. Not introducing any components but kept in mind to be autowired while doing BPS-919 (link above). Saying that this is a helper PR to set the ground for further development. Fully unit tested and proven logging is there. Further components injected in those processors (and possibly task) can be mocked and verified in existing unit tests accompanied with new ones. Plus integration tests later on once fully autowired.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.List;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class ContainerProcessor {
+
+    private static final Logger LOGGER = getLogger(ContainerProcessor.class);
+
+    private static final BlobLeaseClientBuilder LEASE_CLIENT_BUILDER = new BlobLeaseClientBuilder();
```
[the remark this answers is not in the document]
doncem: Should be. Building a client (calling method `buildAsyncClient`) creates a new client every time. Blob clients are bounded per blob item so have to build new client every time as well.

src/main/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientManager.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    private static boolean isContainerEnabled(
+        Map<String, ServiceConfiguration.StorageConfig> storageConfiguration,
+        String containerName
+    ) {
+        if (storageConfiguration.containsKey(containerName)) {
+            return storageConfiguration.get(containerName).isEnabled();
+        } else {
+            throw new ServiceConfigNotFoundException("No service configuration found for " + containerName);
```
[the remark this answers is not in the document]
doncem: Service configuration is key component for service and is tight dependency on containers which are already configured/created. Each service/container must have a mapping

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/StorageContainerTask.java
```diff
@@ -0,0 +1,42 @@
...
+import com.azure.storage.blob.BlobServiceAsyncClient;
+import com.azure.storage.blob.models.BlobContainerItem;
+import org.slf4j.Logger;
+import uk.gov.hmcts.reform.blobrouter.config.ServiceConfiguration;
+import uk.gov.hmcts.reform.blobrouter.tasks.processors.ContainerProcessor;
+import uk.gov.hmcts.reform.blobrouter.util.StorageClientManager;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class StorageContainerTask {
```
lgonczar: Worth renaming, so that the name of the class shows the purpose of the task.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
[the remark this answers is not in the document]
rkondratowicz: I think it's not, based on the nature of builder pattern.
It's mutable and calling it may contain 'remains' from previous call.

Like here:
```java
var abc = BUILDER.setA("abc").setB(100).build()
...
var xyz = BUILDER.setB(100).build();
```
`xyz` now has also property `A` set to `"abc"`.

Please init builder for every usage.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/StorageContainerTask.java
```diff
@@ -0,0 +1,42 @@
...
+        ContainerProcessor containerProcessor,
+        ServiceConfiguration serviceConfiguration
+    ) {
+        this.containerProcessor = containerProcessor;
+        this.serviceConfiguration = serviceConfiguration;
+        this.storageClient = storageClient;
+    }
+
+    public void run() {
+        LOGGER.info("Started job");
```
lgonczar: I know that we'll have logger name, but it would still be worth mentioning what kind of job we started (and finished).

lgonczar:
https://tools.hmcts.net/jira/browse/BPS-914 is done and https://tools.hmcts.net/jira/browse/BPS-919 is not to be started before all the stories from current sprint are finished. Please pause work in this one.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobRouterMainTask.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public void run() {
+        LOGGER.info("Started {} job", TASK_NAME);
+
+        StorageClientManager
+            .getAvailableContainers(storageClient, serviceConfiguration)
+            .toStream()
+            .map(BlobContainerItem::getName)
+            .map(storageClient::getBlobContainerAsyncClient)
```
lgonczar: Given we're just retrieving blob container client in order to immediately pass it to `containerProcessor`, I think it would be better to move `storageClient` to `ContainerProcessor` class from here, so that `containerProcessor` can get the right client itself.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+        LOGGER.info("Processing container {}", containerName);
+
+        containerClient
+            .listBlobs()
+            .subscribe(
+                blob -> {
+                    processBlobs(containerClient, blob);
+                    atomInt.updateAndGet(operand -> ++operand);
+                },
+                null, // error consumer. can be implemented later
```
lgonczar: We don't need to bring the whole job in at once, but when we're adding a class that's responsible for some task, that class needs to do its thing properly. Error handling is quite a crucial element of any processing and it can't be added later, as if it was additional functionality.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+                () -> LOGGER.info("Finished processing container {}. Blobs found: {}", containerName, atomInt.get())
+            );
+    }
+
+    private void processBlobs(BlobContainerAsyncClient containerClient, BlobItem blob) {
```
lgonczar: `processBlob`?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final BlobProcessor blobProcessor;
+
+    public ContainerProcessor(BlobProcessor blobProcessor) {
+        this.blobProcessor = blobProcessor;
+    }
+
+    public void process(BlobContainerAsyncClient containerClient) {
+        String containerName = containerClient.getBlobContainerName();
+        AtomicInteger atomInt = new AtomicInteger(0);
```
lgonczar: The name of the variable should reflect its purpose, e.g.  `processedBlobCount`.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+
+        containerClient
+            .listBlobs()
+            .subscribe(
+                blob -> {
+                    processBlobs(containerClient, blob);
+                    atomInt.updateAndGet(operand -> ++operand);
+                },
+                null, // error consumer. can be implemented later
+                () -> LOGGER.info("Finished processing container {}. Blobs found: {}", containerName, atomInt.get())
```
lgonczar: Processing a blob may not be completed because of an error or because there was already a lease on it. The number of blobs in the container is useful information, but the number of blobs we actually processed is, in my opinion, the most important piece to include.

build.gradle
```diff
@@ -204,7 +204,7 @@ ext.libraries = [
 ext["rest-assured.version"] = '4.1.2'
 
 dependencies {
-  implementation group: 'com.azure', name: 'azure-storage-blob', version: '12.1.0'
+  implementation group: 'com.azure', name: 'azure-storage-blob', version: '12.0.1'
```
rkondratowicz: why the downgrade?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+        AtomicInteger atomInt = new AtomicInteger(0);
+
+        LOGGER.info("Processing container {}", containerName);
+
+        containerClient
+            .listBlobs()
+            .subscribe(
+                blob -> {
+                    processBlobs(containerClient, blob);
+                    atomInt.updateAndGet(operand -> ++operand);
```
rkondratowicz: there's also `.incrementAndGet()`

PR #49, opened 2019-12-13
Remove environment variables from application config

doncem:
### Change description ###

Do not require env var for storage secrets and let spring bootstrap to set them later before validating the configuration file

```
[ ] Yes
[x] No
```


doncem:
not needed. actual problem discovered. fix: #51

PR #50, opened 2019-12-13
Add functional test - upload and delete blob

doncem:

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/UploadAndDeleteBlobTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static uk.gov.hmcts.reform.blobrouter.util.StorageHelper.CONTAINER_NAME;
+import static uk.gov.hmcts.reform.blobrouter.util.StorageHelper.GET_BLOB_CLIENT;
+
+class UploadAndDeleteBlobTest {
+
+    private static final String TEST_URL = CONFIG.getString("test-url");
+
+    @DisplayName("Upload using library client and delete afterwards")
+    @Test
+    void should_upload_and_delete_using_client() {
```
rkondratowicz: this test does not test the application

doncem:
No beneficial value here right now. When actual job will run in aat environment it'll consume blobs and this test might be fluctuating. We will/should introduce functional tests once actual purpose of service is introduced (even partially is enough)

Will try to keep in mind gradle config to be applied later

PR #62, opened 2020-01-03
Copy bulk scan storage account details from bulk scan to reform scan key vault

aliveni:

### Change description ###
Copy bulk scan storage account details from bulk-scan key vault to refomr-scan key vault.
These properties will be used in the blob processor scheduled task, which will be introduced in the next PR.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -10,8 +10,8 @@ java:
         - storage-account-name
         - storage-account-primary-key
         - storage-account-secondary-key
-        - reform-bulk-scan-storage-account-name
-        - reform-bulk-scan-storage-account-primary-key
+        - bulkscan-storage-account-name
+        - bulkscan-storage-account-primary-key
   environment:
   image: 'hmctspublic.azurecr.io/reform-scan/blob-router:latest'
 blobstorage:
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -10,6 +10,8 @@ java:
         - storage-account-name
         - storage-account-primary-key
         - storage-account-secondary-key
+        - bulk-scan-storage-account-name
+        - bulk-scan-storage-account-primary-key
   environment:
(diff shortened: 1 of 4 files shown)


infrastructure/main.tf
```diff
@@ -1,3 +1,41 @@
 provider "azurerm" {
   version = "=1.33.1"
 }
+
+locals {
+  reform-scan-vault-name = "reform-scan-${var.env}"
+  bulk-scan-vault-name = "bulk-scan-${var.env}"
+}
+
+data "azurerm_key_vault" "bulk-scan_key_vault" {
```
doncem: let's use terraform specific references all with underscore

charts/reform-scan-blob-router/values.yaml
```diff
@@ -10,6 +10,8 @@ java:
         - storage-account-name
         - storage-account-primary-key
         - storage-account-secondary-key
+        - bulk-scan-storage-account-name
+        - bulk-scan-storage-account-primary-key
```
aliveni: if the keys with `cft` prefix are preferable, I'm happy to rename them.

infrastructure/main.tf
```diff
@@ -1,3 +1,41 @@
...
+  name                = "${local.bulk-scan-vault-name}"
+  resource_group_name = "bulk-scan-${var.env}"
+}
+
+data "azurerm_key_vault_secret" "cft_storage_account_name" {
```
doncem: perhaps we don't need to know about the abbreviations here? shall we just prefix with `bulk_scan` instead?

charts/reform-scan-blob-router/values.yaml
nitinprabhu: There is no change in `bootstrap.yaml` so not sure adding it here and we are setting that value to ? Also if we update `values.yaml` we need to bump chart version. I would say we don't need any change to this file at this point and best to update this when you use it.

doncem: ah. yes, i refrained from chart jiggery here..

PR #65, opened 2020-01-06
add OSBA blob storage for preview

mdayican:

### Change description ###

adding OSBA blob storage for preview

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/Chart.yaml
+++ b/charts/reform-scan-blob-router/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for Blob Router Service
 name: reform-scan-blob-router
 home: https://github.com/hmcts/blob-router-service
-version: 0.0.4
+version: 0.0.5
 maintainers:
   - name: HMCTS BSP Team
     email: bspteam@hmcts.net
--- a/charts/reform-scan-blob-router/values.preview.template.yaml
+++ b/charts/reform-scan-blob-router/values.preview.template.yaml
@@ -31,3 +31,14 @@ reformblobstorage:
     containers:
       - bulkscan
       - bulkscan-rejected
+
+cftblobstorage:
+  enabled: true
+  resourceGroup: reform-scan-aks
+  releaseNameOverride: cft-storage-secret-{{ .Release.Name }}
+  teamName: "BSP"
+  location: ukwest
+  setup:
+    containers:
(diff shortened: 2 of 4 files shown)


charts/reform-scan-blob-router/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for Blob Router Service
 name: reform-scan-blob-router
 home: https://github.com/hmcts/blob-router-service
-version: 0.0.4
+version: 0.0.5
```
doncem: Can we firstly merge flux config PR?

charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -9,18 +9,36 @@ java:
...
       secretRef: storage-secret-{{ .Release.Name }}
       key: accessKey
+    STORAGE_BULKSCAN_ACCOUNT_NAME:
+      secretRef: storage-secret-cft-{{ .Release.Name }}
+      key: storageAccountName
+    STORAGE_BULKSCAN_ACCOUNT_KEY:
+      secretRef: storage-secret-cft-{{ .Release.Name }}
+      key: accessKey
   environment:
+    SPRING_CLOUD_PROPERTIESVOLUME_ENABLED: false
```
nitinprabhu: why do we need this change ?

mdayican: otherwise it bootstraps  from secret and overrides osba connections

charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -9,18 +9,36 @@ java:
...
-blobstorage:
+reformblobstorage:
+  enabled: true
   resourceGroup: reform-scan-aks
   teamName: "BSP"
   location: ukwest
   setup:
     containers:
       - bulkscan
       - bulkscan-rejected
```
nitinprabhu: what about crime container ?

charts/reform-scan-blob-router/values.preview.template.yaml
nitinprabhu: as discussed please add a comment here as this needs to be removed once we have key vault config

mdayican: added

PR #66, opened 2020-01-07
Feature/add blob dispatcher task

aliveni:

### Change description ###
Added Blob Dispatching task. 
Code for copying blobs from reform storage account to bulk scan storage account will be added in the next pr.
This code is extracted from #46

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class ContainerProcessor {
+
+    private static final Logger LOGGER = getLogger(ContainerProcessor.class);
+
+    public void process(BlobContainerAsyncClient containerClient) {
+        String containerName = containerClient.getBlobContainerName();
+        AtomicInteger processedBlobCount = new AtomicInteger(0);
```
mdayican: how this will work if there will be  multiple instance?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
```diff
@@ -0,0 +1,44 @@
...
+import org.slf4j.Logger;
+import uk.gov.hmcts.reform.blobrouter.config.ServiceConfiguration;
+import uk.gov.hmcts.reform.blobrouter.tasks.processors.ContainerProcessor;
+import uk.gov.hmcts.reform.blobrouter.util.StorageClientManager;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class BlobDispatcherTask {
+
+    static final String TASK_NAME = "blob-dispatcher";
```
mdayican: access modifier missing
and do we really need this, without this logging may be faster ?

build.gradle
```diff
@@ -208,7 +208,7 @@ ext.libraries = [
 ext["rest-assured.version"] = '4.1.2'
 
 dependencies {
-  implementation group: 'com.azure', name: 'azure-storage-blob', version: '12.1.0'
+  implementation group: 'com.azure', name: 'azure-storage-blob', version: '12.0.1'
```
rkondratowicz: why the downgrade?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
```diff
@@ -0,0 +1,44 @@
...
+    private final ServiceConfiguration serviceConfiguration;
+
+    public BlobDispatcherTask(
+        BlobServiceAsyncClient storageClient,
+        ContainerProcessor containerProcessor,
+        ServiceConfiguration serviceConfiguration
+    ) {
+        this.containerProcessor = containerProcessor;
+        this.serviceConfiguration = serviceConfiguration;
+        this.storageClient = storageClient;
```
rkondratowicz: would be nice to keep the parameters order

src/main/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientManager.java
```diff
@@ -0,0 +1,39 @@
...
+        // utility class constructor
+    }
+
+    public static Flux<BlobContainerItem> getAvailableContainers(
+        BlobServiceAsyncClient storageClient,
+        ServiceConfiguration serviceConfiguration
+    ) {
+        return storageClient
+            .listBlobContainers()
+            .filter(item -> !item.getName().endsWith(REJECTED_CONTAINER_PREFIX))
```
rkondratowicz: if it ends with it, it's a **suffix**

src/main/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientManager.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    private static boolean isContainerEnabled(
+        Map<String, ServiceConfiguration.StorageConfig> storageConfiguration,
+        String containerName
+    ) {
+        if (storageConfiguration.containsKey(containerName)) {
+            return storageConfiguration.get(containerName).isEnabled();
+        } else {
+            throw new ServiceConfigNotFoundException("No service configuration found for " + containerName);
```
rkondratowicz: Looks like this would prevent all other containers to be processed.
Perhaps we could log an error and return `false` here instead?

rkondratowicz: "No service configuration found for hello" does not make it clear it's a container.

src/test/resources/storage/list-containers.json
```diff
@@ -0,0 +1,11 @@
...
+  "Containers": [{
+    "Name": "bulkscan"
+  }, {
+    "Name": "bulkscan-rejected"
+  }, {
+    "Name": "disabled-service"
+  }, {
+    "Name": "disabled-service-rejected"
+  }]
+}
```
rkondratowicz: please format

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
```diff
@@ -0,0 +1,30 @@
...
+        String query = request.getUrl().getQuery();
+
+        if (LIST_CONTAINERS.equals(path + "?" + query)) {
+            return Mono.just(
+                new MockHttpResponse(request, 200, getFileContents("storage/list-containers.json"))
+            );
+        }
+        Assertions.fail("Request '" + request.getUrl() + "' is not setup");
+
+        return Mono.empty();
```
rkondratowicz: could be wrapped in `else`

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientsHelper.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public static void setAzureTestMode() {
+        Configuration.getGlobalConfiguration().put("AZURE_TEST_MODE", TestMode.RECORD.name());
+    }
+
+    public static BlobServiceAsyncClient getStorageClient(InterceptorManager interceptorManager) {
+        return STORAGE_CLIENT_BUILDER
+            .credential(STORAGE_CREDENTIALS)
+            .endpoint("http://httpbin.org")
```
rkondratowicz: deserves a comment in code

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
```diff
@@ -0,0 +1,30 @@
...
+    public Mono<HttpResponse> send(HttpRequest request) {
+        String path = request.getUrl().getPath();
+        String query = request.getUrl().getQuery();
+
+        if (LIST_CONTAINERS.equals(path + "?" + query)) {
+            return Mono.just(
+                new MockHttpResponse(request, 200, getFileContents("storage/list-containers.json"))
+            );
+        }
+        Assertions.fail("Request '" + request.getUrl() + "' is not setup");
```
rkondratowicz: 'set up'

build.gradle
aliveni: the `azure-core-test` library is not available for `12.1.0` version

aliveni: the azure-core-test library is not available for `12.1.0` version

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
rkondratowicz: still fine in the context of single instance.

mdayican: it may print same number many times?

rkondratowicz: each instance will process different files and log how many files this particular instance has processed

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.azure.storage.blob.BlobContainerAsyncClient;
+import com.azure.storage.blob.models.BlobItem;
+import org.slf4j.Logger;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+@SuppressWarnings("java:S1135")
```
rkondratowicz: it would be nice to add a comment

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
mdayican: when pod crashed new pod starts from zero, what is the point?

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessorTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private ContainerProcessor containerProcessor;
+
+    @Override
+    protected void beforeTest() {
+        containerProcessor = new ContainerProcessor();
+    }
+
+    @Test
+    @SuppressWarnings("java:S2925")
```
rkondratowicz: what's this one about?

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
```diff
@@ -0,0 +1,30 @@
...
+class StorageStubHttpClient implements HttpClient {
+
+    private static final String LIST_CONTAINERS = "?comp=list";
+
+    @Override
+    public Mono<HttpResponse> send(HttpRequest request) {
+        String path = request.getUrl().getPath();
+        String query = request.getUrl().getQuery();
+
+        if (LIST_CONTAINERS.equals(path + "?" + query)) {
```
rkondratowicz: so `path` has to be empty?

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientsHelper.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public static void setAzureTestMode() {
+        Configuration.getGlobalConfiguration().put("AZURE_TEST_MODE", TestMode.RECORD.name());
+    }
+
+    public static BlobServiceAsyncClient getStorageClient(InterceptorManager interceptorManager) {
+        return STORAGE_CLIENT_BUILDER
+            .credential(STORAGE_CREDENTIALS)
+            .endpoint("http://httpbin.org") // HTTP request and response service
```
rkondratowicz: will unit tests call this service?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+import org.slf4j.Logger;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+@SuppressWarnings("java:S1135") // ignore TODOs. will be removed when blob processing part is implemented
+public class ContainerProcessor {
+
+    private static final Logger LOGGER = getLogger(ContainerProcessor.class);
```
lgonczar: We typically name loggers `logger`. Maybe we should stick to it? It's static final, but it doesn't represent a value.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+        LOGGER.info("Processing container {}", containerName);
+
+        containerClient
+            .listBlobs()
+            .subscribe(
+                blob -> {
+                    processBlob(blob);
+                    processedBlobCount.incrementAndGet();
+                },
+                null, // TODO: error consumer
```
lgonczar: We have to handle errors at this point, at least by logging them properly.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+                () -> LOGGER.info(
+                    "Finished processing container {}. Blobs processed: {}", containerName, processedBlobCount.get()
```
lgonczar: `Blobs processed` is misleading, because we may have skipped some of those blobs (the ones with leases acquired). Maybe we should increment that counter conditionally, only when we actually processed the blob? Alternatively, we can remove this counter from this PR and add it later.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+                )
+            );
+    }
+
+    @SuppressWarnings("java:S1172")
+    // ignore unused variable. This will be removed once the blob processor code is implemented
+    private void processBlob(BlobItem blob) {
+        // TODO: process blob
```
lgonczar: This change should be done before this PR, i.e. dependencies should be ready first, rather than having some orchestration that doesn't do anything. Otherwise, we have a feature, but we don't have proper tests for it, because it doesn't work yet. Then we add all necessary dependencies, write all tests for them and consider the story done, without realising that there are tests missing at higher level.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+        AtomicInteger processedBlobCount = new AtomicInteger(0);
+
+        LOGGER.info("Processing container {}", containerName);
+
+        containerClient
+            .listBlobs()
+            .subscribe(
+                blob -> {
+                    processBlob(blob);
+                    processedBlobCount.incrementAndGet();
```
rkondratowicz: Blob may not be processed due to not job not being able to acquire a lock, but the counter will be increased here anyway. Not what we want I assume.

src/main/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientManager.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private static final Logger LOGGER = getLogger(StorageClientManager.class);
+
+    private static final String REJECTED_CONTAINER_SUFFIX = "rejected";
+
+    private StorageClientManager() {
+        // utility class constructor
+    }
+
+    public static Flux<BlobContainerItem> getAvailableContainers(
```
lgonczar: For better testability, could this method be non-static?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
aliveni: looks ok in IntelliJ

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessorTest.java
aliveni: it's about calling Thread.sleep() method in the tests. Added a comment now

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
aliveni: yes, for list of containers it is empty

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientsHelper.java
aliveni: no, but the Azure test library is validating the URL with this exact value

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public void run() {
+        LOGGER.info("Started {} job", TASK_NAME);
+
+        StorageClientManager
+            .getAvailableContainers(storageClient, serviceConfiguration)
+            .toStream()
+            .map(BlobContainerItem::getName)
+            .map(storageClient::getBlobContainerAsyncClient)
+            .forEach(containerProcessor::process);
```
lgonczar: What we're doing here is:
- get existing containers, but only those that are enabled in configuration
- for each of those containers get a client
- using the appropriate client, process each container

While technically it does the job, I'd recommend an alternative approach:
- get the names of all enabled containers in configuration
- for each of those names, process the container of a given name

There are two benefits for the second approach:
- When the container from the configuration doesn't exist, we'll get an error. In current approach, we don't know about the problem.
- This task can just use a `containerProcessor` and not `storageClient`. `containerProcessor` could use `storageClient` directly, and these lower-level details could be hidden from the task. That would make it easier to test.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
lgonczar: Aaa, sorry, what Rafał said.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
lgonczar: This counter is for a single job run. The purpose is to know how many files were processed within that job run.

src/main/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientManager.java
lgonczar: @mdayican your argument?

mdayican: if you are telling this for mocking we can still mock static methods.

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientsHelper.java
rkondratowicz: surely this deserves a comment then. I'd expect this to normally be something like localhost

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
rkondratowicz: I mean that parameters are in this order:
- storageClient
- containerProcessor
- serviceConfiguration

But are then used in a different order:
- containerProcessor
- serviceConfiguration
- storageClient

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
aliveni: done

src/main/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientManager.java
lgonczar: Mocking static methods is not quite as straightforward as instance mocks and has its own limitations. There's no cost in having instance methods.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
aliveni: This variable is being used in the test. It may be a tiny bit faster but we don't have to hard code the task name in the verifications

src/test/java/uk/gov/hmcts/reform/blobrouter/util/StorageClientsHelper.java
aliveni: added comment

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
aliveni: Updated the parameters order

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
aliveni: added logging for now

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
aliveni: Removed the counter

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
aliveni: updated the log message. `processedBlobCount` variable is removed

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
aliveni: Updated the pr as suggested

PR #68, opened 2020-01-07
Copy error notification secrets

doncem:

[Notification Client for blob router](https://tools.hmcts.net/jira/browse/BPS-966)

### Change description ###

Before being able to proceed any further - we need secrets to be present in reform scan key vault. Copying them from already existing bulk scan ones

```
[ ] Yes
[x] No
```


doncem:
> can this go in reform shared infra ?

it can but followed existing copying present in repo

PR #69, opened 2020-01-08
Replace defaults with actual values

doncem:

[Notification Client for blob router](https://tools.hmcts.net/jira/browse/BPS-966)

### Change description ###

Secrets are copied during #68. About to create flux config PR referencing this one

```
[ ] Yes
[x] No
```


doncem:
Merging as service is not alive yet - no need to monitor even though I will do

PR #71, opened 2020-01-08
db infra

rkondratowicz:


infrastructure/main.tf
```diff
@@ -7,6 +7,20 @@ locals {
   bulk-scan-vault-name   = "bulk-scan-${var.env}"
 }
 
+module "blob-router-db" {
```
doncem: It is just the `db`. Prefix is over-informing the fact

infrastructure/main.tf
```diff
@@ -91,3 +105,42 @@ resource "azurerm_key_vault_secret" "error_notifications_username" {
...
+  key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
+  value        = "${module.blob-router-db.postgresql_listen_port}"
+}
+
+resource "azurerm_key_vault_secret" "POSTGRES_DATABASE" {
+  name         = "${var.component}-POSTGRES-DATABASE"
+  key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
+  value        = "${module.blob-router-db.postgresql_database}"
+}
+
```
doncem: Would be nice to not copy paste legacy `resources` because somebody in the past put down these values. Would be nice to expect next PR amending these points:

- resource names lowercase + underscores
- key vault names lowercase + hyphens

P.S. no need to reference `POSTGRESS`. It doesn't really matter which DB we use. It matters if component provides customisation to public. Since it doesn't - no matter.

P.S. name prefix I believe should just be `db`. If so, the reference to `postgres` could stay. So we could do as follows:

```
name = "db-postgres-user"
...
```

infrastructure/main.tf
```diff
@@ -91,3 +105,42 @@ resource "azurerm_key_vault_secret" "error_notifications_username" {
...
+# Copy postgres password for flyway migration
+resource "azurerm_key_vault_secret" "flyway_password" {
+  name         = "flyway-password"
+  value        = "${module.blob-router-db.postgresql_password}"
+  key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
+}
+# endregion
```
doncem: If it is just a copy - there's not need for it. Configuration can directly use original value

PR #77, opened 2020-01-09
Add crime storage config

mdayican:

### Change description ###

Add crime storage config for preview and add bootstrap from vault.
Crime storage connection details are moved bulk scan aat vault to reform blob router vault.

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/requirements.yaml
```diff
@@ -12,3 +12,8 @@ dependencies:
     version: ~0.2.1
     repository: '@hmctspublic'
     condition: cftblobstorage.enabled
+  - name: blobstorage
+    alias: crimeblobstorage
+    version: ~0.2.1
+    repository: '@hmctspublic'
+    condition: cftblobstorage.enabled
```
lgonczar: We have three blob storage charts in dependencies - reform, cft and crime. Do we need all three? Also, should the condition be different for crime?

mdayican: we want to create 3 different blob storage so we can test moving from one blob storage to another one, for helm alias is only way for now

lgonczar: I got the three blob storage charts now, just the question about the condition

charts/reform-scan-blob-router/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for Blob Router Service
 name: reform-scan-blob-router
 home: https://github.com/hmcts/blob-router-service
-version: 0.0.6
+version: 0.0.7
```
nitinprabhu: don't forget to update flux :)

src/main/resources/bootstrap.yaml
```diff
@@ -15,3 +15,5 @@ spring:
         storage-account-secondary-key:  storage.account-secondary-key
         bulkscan-storage-account-name: storage.bulkscan.account-name
         bulkscan-storage-account-primary-key:  storage.bulkscan.account-key
+        crime-storage-account-name: storage.crime.account-name
+        crime-storage-account-primary-key: storage.crime.account-key
```
nitinprabhu: you have not added this to preview key vault configuration ?

PR #78, opened 2020-01-09
Add flyway and first migration

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-982


charts/reform-scan-blob-router/values.yaml
```diff
@@ -15,7 +15,11 @@ java:
         - storage-account-secondary-key
         - bulkscan-storage-account-name
         - bulkscan-storage-account-primary-key
+        - flyway-password
   environment:
+    FLYWAY_URL: jdbc:postgresql://blob_router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/bulk_scan?sslmode=require
+    FLYWAY_USER: blob_router@blob_router-{{ .Values.global.environment }}
+    FLYWAY_SKIP_MIGRATIONS: true
```
rkondratowicz: thought it would be clearer than `FLYWAY_NOOP_STRATEGY`

src/main/resources/application.yaml
```diff
@@ -30,3 +30,6 @@ service:
       sas-validity: 300 #In seconds
     - name: sscs
       sas-validity: 300 #In seconds
+
+flyway:
+  skip-migrations: ${FLYWAY_SKIP_MIGRATIONS:false}
```
nitinprabhu: best to avoid having default values as it makes it easy if containers don't come and fail

charts/reform-scan-blob-router/values.yaml
nitinprabhu: this value should it be false for preview ?

rkondratowicz: updated preview template file

src/main/java/uk/gov/hmcts/reform/blobrouter/config/FlywayConfig.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.hmcts.reform.blobrouter.config;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.flyway.FlywayMigrationStrategy;
+import org.springframework.context.annotation.Bean;
+
+import java.util.stream.Stream;
+
+public class FlywayConfig {
```
doncem: Might as well be super safe by adding following:

```java
@AutoConfigureAfter({
    DataSourceAutoConfiguration.class
})
@AutoConfigureBefore({
    FlywayAutoConfiguration.class
})
@Configuration
@ConditionalOnClass(Flyway.class)
@ConditionalOnProperty(prefix = "spring.flyway", name = "enabled", matchIfMissing = true)
```

src/main/resources/application.yaml
nitinprabhu: can we please remove default value ?

rkondratowicz: kk

src/main/java/uk/gov/hmcts/reform/blobrouter/config/FlywayConfig.java
rkondratowicz: done

nitinprabhu:
> do not you need postgresql setting for Preview?

good point @mdayican Postgres config for preview should go first and then this PR. I missed this assuming it is already there for preview

Jenkinsfile_CNP
```diff
@@ -42,6 +42,7 @@ withPipeline(type, product, component) {
     builder.gradle('integration')
   }
 
+  enableDbMigration(dbMigrationVaultName = 'reform-scan')
```
nitinprabhu: it should be `enableDbMigration('reform-scan')` or you sure current one works ?

rkondratowicz: ?

PR #79, opened 2020-01-09
Add blob dispatcher component

doncem:

### Change description ###

Adding component for uploading zip files to storage containers. Including only one dependency which is not yet beaned up. Current plan we have: 2 destinations meaning 2 autowired storage clients

~P.S. Need advice/help on integration test~

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcherTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcherTest.java
@@ -12,6 +12,7 @@ class BlobDispatcherTest extends TestBase {
 
     private static final String NEW_BLOB_NAME = "new.blob";
     private static final String DESTINATION_CONTAINER = "bulkscan";
+    private static final String BOGUS_CONTAINER = "bogus";
 
     private BlobDispatcher dispatcher;
 
@@ -32,6 +33,13 @@ class BlobDispatcherTest extends TestBase {
     void should_upload_blob_to_dedicated_container() {
         assertThatCode(() -> dispatcher
             .dispatch(NEW_BLOB_NAME, NEW_BLOB_NAME.getBytes(), DESTINATION_CONTAINER)
+        ).doesNotThrowAnyException();
+    }
+
+    @Test
+    void should_catch_BlobStorageException_and_suppress_it() {
+        assertThatCode(() -> dispatcher
+            .dispatch(NEW_BLOB_NAME, NEW_BLOB_NAME.getBytes(), BOGUS_CONTAINER)
         ).doesNotThrowAnyException();
     }
 }
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcherTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcherTest.java
@@ -1,12 +1,10 @@
 package uk.gov.hmcts.reform.blobrouter.services.storage;
 
 import com.azure.core.test.TestBase;
-import com.azure.storage.blob.BlobServiceAsyncClient;
+import com.azure.storage.blob.BlobServiceClient;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import uk.gov.hmcts.reform.blobrouter.util.StorageClientsHelper;
-
-import java.time.Duration;
 
 import static org.assertj.core.api.Assertions.assertThatCode;
 
@@ -26,7 +24,7 @@ class BlobDispatcherTest extends TestBase {
 
     @Override
     protected void beforeTest() {
-        BlobServiceAsyncClient storageClient = StorageClientsHelper.getStorageClient(interceptorManager);
+        BlobServiceClient storageClient = StorageClientsHelper.getStorageClient(interceptorManager);
         dispatcher = new BlobDispatcher(storageClient);
     }
 
@@ -34,7 +32,6 @@ class BlobDispatcherTest extends TestBase {
     void should_upload_blob_to_dedicated_container() {
         assertThatCode(() -> dispatcher
             .dispatch(NEW_BLOB_NAME, NEW_BLOB_NAME.getBytes(), DESTINATION_CONTAINER)
-            .block(Duration.ofSeconds(2)) // max waiting time
         ).doesNotThrowAnyException();
     }
 }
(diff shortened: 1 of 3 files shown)


src/integrationTest/resources/storage/upload-blob.json
```diff
@@ -0,0 +1,12 @@
...
+  "Server" : "Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0",
+  "ETag" : "0x8D761B5AF5CA061",
+  "Last-Modified" : "Tue, 05 Nov 2019 06:02:05 GMT",
+  "retry-after" : "0",
+  "Content-Length" : "0",
+  "StatusCode" : "201",
+  "x-ms-request-id" : "bf1765bd-101e-0029-779e-934634000000",
+  "Date" : "Tue, 05 Nov 2019 06:02:05 GMT",
+  "x-ms-client-request-id" : "986d270c-9648-48da-8e7d-0261d486b848"
+}
```
doncem: cheekily copied from azure test playback

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static final Logger logger = getLogger(BlobDispatcher.class);
+
+    private final BlobServiceAsyncClient storageClient;
+
+    public BlobDispatcher(BlobServiceAsyncClient storageClient) {
+        this.storageClient = storageClient;
+    }
+
+    public void dispatch(String blobName, ZipInputStream inputStream, String destinationContainer) throws IOException {
```
rkondratowicz: will the caller now what to know whether this operation succeeded?

doncem: There's `onSuccess` method which can do finalisation steps. The caller should not amend anything related to this processing route. 2 notes:

- Lease should expire automatically after 60s (max value)
- Deletion is dedicated to separate job. If not, we need to orchestrate all these calls differently by removing subscription calls and returning flux/mono to continue composition on caller side. Could also improve testing framework by allowing test framework's subscriber to watch all this

lgonczar: If the upload succeeds, the caller will need to save the record in the DB. In either, success or failure, it should break the lease, as it's done working on the blob. We could theoretically pass logic to be executed on success/failure, but that would be an inelegant solution, I think.

lgonczar: Any thoughts about the post-update action?

rkondratowicz: Then suddenly this class will also need a dependency on repository layer.
I thought whatever will be calling this method will then handle the next step / call class handling the next step.
Alternatively, an action that should be called after success could be passed to this method as parameter.

Regardless, worth adding unit test.

doncem: After posting my comment I am more leaning towards composing all actions. It'll help test and apply steps

lgonczar: I'm not that much agains passing logic - we can see. But I think we should pass a byte array into this method so that:
- this method doesn't synchronously load the blob before asynchronously uploading it
- we don't have any zip-specific failure in this method - at this point we know we want to upload it and verification should happen earlier

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
```diff
@@ -0,0 +1,49 @@
...
+        logger.info("Uploading {} to {} container", blobName, destinationContainer);
+
+        try {
+            getContainerClient(destinationContainer)
+                .getBlobClient(blobName)
+                .getBlockBlobClient()
+                .upload(
+                    new ByteArrayInputStream(blobContents),
+                    blobContents.length,
+                    false // overwrite flag
```
nitinprabhu: If we are not overwriting then it will create snapshots ?

We are not handling snapshots so may be overwrite it for now ?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
```diff
@@ -1,19 +1,24 @@
...
+import java.util.Map;
+
 import static uk.gov.hmcts.reform.blobrouter.util.ResourceFilesHelper.getFileContents;
 
 class StorageStubHttpClient implements HttpClient {
 
     private static final String LIST_CONTAINERS = "?comp=list";
     private static final String LIST_ONE_BLOB = "/bulkscan?include=&restype=container&comp=list";
     private static final String LIST_ZERO_BLOBS = "/empty?include=&restype=container&comp=list";
+    private static final String UPLOAD_NEW_BLOB = "/bulkscan/new.blob?null";
```
nitinprabhu: null ? Could you please explain ?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
doncem: i can remove the flag addition

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
doncem: not sure. when running without null test informs me this particular url is not configured. strange serialisation 🤷‍♂

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcherTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.azure.core.test.TestBase;
+import com.azure.storage.blob.BlobServiceClient;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import uk.gov.hmcts.reform.blobrouter.util.StorageClientsHelper;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+class BlobDispatcherTest extends TestBase {
```
nitinprabhu: Test missing for exception scenario ?

nitinprabhu: Just logging in exception never mind

doncem: Added anyway. More complete coverage and in case this edge case will change with like error notification ping - will have to be addressed

PR #80, opened 2020-01-10
Add subscriber for processing blobs

doncem:

### Change description ###

The main composer class which step-by-step processes single blob. Blob clients are strictly per blob so each time this class will have to constructed - not beaned up.

Should contain error handler in the future so that only serious errors are logged and therefore - break the chain of blob processing from the container

```
[ ] Yes
[x] No
```


lgonczar:
We need a discussion in a team - I'd like to see how this class would be used in our code

doncem:
Change of development course

PR #82, opened 2020-01-10
Create new model for envelope inserts

rkondratowicz:
the write model does not have the following fields:
- ID (to be generated when inserting to DB)
- createdAt (to be generated when inserting to DB)
- isDeleted (always false when inserting to DB)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/NewEnvelope.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/NewEnvelope.java
@@ -1,0 +1,20 @@
+package uk.gov.hmcts.reform.blobrouter.data.model;
+
+import java.time.Instant;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/NewEnvelope.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.hmcts.reform.blobrouter.data.model;
+
+import java.time.Instant;
+
+public class NewEnvelope {
```
nitinprabhu: can we not call it as just Envelope ?

rkondratowicz: there's already one. for reads

PR #83, opened 2020-01-13
wait blob processing config

mdayican:

https://tools.hmcts.net/jira/browse/BPS-801

### Change description ###

add blob processing wait config 

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -22,8 +22,7 @@ java:
     FLYWAY_URL: jdbc:postgresql://blob_router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/bulk_scan?sslmode=require
     FLYWAY_USER: blob_router@blob_router-{{ .Values.global.environment }}
     FLYWAY_SKIP_MIGRATIONS: true
-    #Change for Prod, wait required time to get blob
-    STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
+    STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 30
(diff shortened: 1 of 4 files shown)


rkondratowicz:
shouldn't `application.yaml`be updated as well?

charts/reform-scan-blob-router/values.yaml
```diff
@@ -22,6 +22,8 @@ java:
     FLYWAY_URL: jdbc:postgresql://blob_router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/bulk_scan?sslmode=require
     FLYWAY_USER: blob_router@blob_router-{{ .Values.global.environment }}
     FLYWAY_SKIP_MIGRATIONS: true
+    #Change for Prod, wait required time to get blob
+    STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
```
rkondratowicz: maybe it would be safer to use actual value here and overwrite it on preview and aat?

nitinprabhu: yeah good suggestion as in case we forget to override it will apply default which is 30 minutes

mdayican: agree, I thought the same but may be overriding in flux would be safer as any changes here would not effect prod.

mdayican:
> shouldn't `application.yaml`be updated as well?

if required it  will be done with code changes

charts/reform-scan-blob-router/values.yaml
nitinprabhu: any changes here will be applied as default unless overriden in flux or env specific  yaml config.

mdayican: that is why I say may be it is better to override in flux prod.

PR #84, opened 2020-01-13
ContainerCleaner

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-970

### Change description ###
Defined ContainerCleaner class with sync BlobContainerClient

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+import uk.gov.hmcts.reform.blobrouter.data.EnvelopeRepository;
+
+import java.util.Map;
+import java.util.UUID;
+
+import static java.util.stream.Collectors.toMap;
+import static org.slf4j.LoggerFactory.getLogger;
+import static uk.gov.hmcts.reform.blobrouter.data.model.Status.DISPATCHED;
+
+@SuppressWarnings("java:S1135") // ignore TODOs. will be removed when blob deletion part is implemented
```
lgonczar: This can go - there are no TODOs

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+            .forEach(blobName -> {
+                tryToDeleteBlob(envelopesToDelete, containerClient, blobName);
+            });
+    }
+
+    private void tryToDeleteBlob(Map<String, UUID> envelopesToDelete, BlobContainerClient containerClient, String blobName) {
+        BlobClient blob = containerClient.getBlobClient(blobName);
+
+        if (envelopesToDelete.containsKey(blobName)) {
+            blob.delete();
```
nitinprabhu: what happens if the delete fails or probably that exception handling is wip ?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+            envelopeRepository.markAsDeleted(envelopesToDelete.get(blobName));
+            logger.info("Deleted file {}", blobName);
```
nitinprabhu: worth adding container name to this log

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+    public ContainerCleaner(
+        BlobServiceClient storageClient,
+        EnvelopeRepository envelopeRepository
+    ) {
+        this.storageClient = storageClient;
+        this.envelopeRepository = envelopeRepository;
+    }
+
+    public void process(String containerName) {
+        logger.info("Deleting from container {}", containerName);
```
lgonczar: Worth restating to say sth like `Deleting processed blobs from container {}`. Up to you.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+
+@SuppressWarnings("java:S1135") // ignore TODOs. will be removed when blob deletion part is implemented
+public class ContainerCleaner {
+
+    private static final Logger logger = getLogger(ContainerCleaner.class);
+
+    private final BlobServiceClient storageClient;
+    private final EnvelopeRepository envelopeRepository;
+
+    public ContainerCleaner(
```
nitinprabhu: `DeleteBlobTask` would be better but current one is also fine

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
lgonczar: I guess we'll have to break the lease first. Otherwise, deletion of the blob might be impossible.

lgonczar: On a second thought, it's fine. If the blob is leased, it's better to skip its processing.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
lgonczar: Worth saying that we deleted a dispatched blob of this name from this container.

lgonczar: We should log something in case of failure to delete given blob.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+        BlobContainerClient containerClient = storageClient.getBlobContainerClient(containerName);
+
+        containerClient
+            .listBlobs()
+            .stream()
+            .map(BlobItem::getName)
+            .forEach(blobName -> {
+                tryToDeleteBlob(envelopesToDelete, containerClient, blobName);
+            });
+    }
```
lgonczar: Worth logging the completion of this process.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public ContainerCleaner(
+        BlobServiceClient storageClient,
+        EnvelopeRepository envelopeRepository
+    ) {
+        this.storageClient = storageClient;
+        this.envelopeRepository = envelopeRepository;
+    }
+
+    public void process(String containerName) {
```
lgonczar: When this method becomes a job, we'll need to add `@SchedulerLock` annotation (and support shedlock in the app), so that there's only one instance of this task.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
lgonczar: This class is likely to become the task class, so best to stick to the convention Nitin mentioned.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
nitinprabhu: good point we would need to break the lease I think as the main job would be continuously running and chances are most of the time they would be leased.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
rkondratowicz: I assume the intention is to have a separate scheduled job class that will call this one

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+        logger.info("Deleting from container {}", containerName);
+
+        Map<String, UUID> envelopesToDelete = envelopeRepository.find(DISPATCHED, false)
+            .stream()
+            .collect(toMap(env -> env.fileName, env -> env.id));
```
rkondratowicz: why not just use collection of envelopes?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+        BlobContainerClient containerClient = storageClient.getBlobContainerClient(containerName);
+
+        containerClient
+            .listBlobs()
+            .stream()
+            .map(BlobItem::getName)
+            .forEach(blobName -> {
```
rkondratowicz: what if blob doesn't exist but there's still entry for it in DB stating deleted == false?

Perhaps it would be better to iterate over entries from DB instead, and if blob doesn't exist, log that fact, and update DB.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    void should_find_blobs_and_delete(CapturedOutput output) {
+        // given
+        createEnvelope("file1.zip");
+        createEnvelope("file2.zip");
+
+        // when
+        containerCleaner.process(CONTAINER_NAME);
+
+        // then
+        assertOutputCapture(output, "file1.zip", "file2.zip");
```
rkondratowicz: check that blob is gone and DB row got updated (`is_deleted` flag)

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessorTest.java
```diff
@@ -1,16 +1,20 @@
...
+import org.springframework.boot.test.system.OutputCaptureExtension;
 import uk.gov.hmcts.reform.blobrouter.util.StorageClientsHelper;
 
-import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThat;
 
 @ExtendWith(MockitoExtension.class)
+@ExtendWith(OutputCaptureExtension.class)
+@SuppressWarnings("java:S2925") // ignore Thread.sleep() call in the tests
 class ContainerProcessorTest extends TestBase {
```
rkondratowicz: should this file be changed in this PR?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,135 @@
...
+                Instant.now(),
+                status
+            );
+            envelopeRepository.insert(envelope);
+        }
+    }
+
+    private void assertFilesIsDeleteState(boolean isDeleted, String... fileNames) {
+        for (String fileName : fileNames) {
+            assertThat(envelopeRepository.find(fileName, CONTAINER_NAME).isPresent()).isTrue();
```
rkondratowicz: batter to run assertion on the optional object (will result in a clearer error message)

rkondratowicz: To be exact:
```java
assertThat(envelopeRepository.find(fileName, CONTAINER_NAME)).hasValueSatisfying(env -> {
    assertThat(env.isDeleted).isEqualTo(isDeleted)
})
```

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,135 @@
...
+                fileName,
+                Instant.now(),
+                Instant.now(),
+                status
+            );
+            envelopeRepository.insert(envelope);
+        }
+    }
+
+    private void assertFilesIsDeleteState(boolean isDeleted, String... fileNames) {
```
rkondratowicz: is -> in
delete -> deleted

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
```diff
@@ -14,28 +15,63 @@
     private static final String LIST_CONTAINERS = "?comp=list";
     private static final String LIST_ONE_BLOB = "/bulkscan?include=&restype=container&comp=list";
     private static final String LIST_ZERO_BLOBS = "/empty?include=&restype=container&comp=list";
+    private static final String FILE_1 = "/bulkscan/file1.zip";
+    private static final String FILE_3 = "/bulkscan/file3.zip";
+    private static final String FILE_CAUSING_404 = "/bulkscan/causes_404.zip";
+    private static final String FILE_CAUSING_500 = "/bulkscan/causes_500.zip";
```
doncem: just a note - while creating and amending this file i have maintained the alphabetical order to those private static declarations. just a not.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/util/StorageStubHttpClient.java
```diff
@@ -14,28 +15,63 @@
...
-                    new MockHttpResponse(request, 200, getFileContents("storage/list-blobs.json"))
-                );
-            case LIST_ZERO_BLOBS:
-                return Mono.just(
-                    new MockHttpResponse(request, 200, getFileContents("storage/list-empty-blobs.json"))
-                );
-            default:
-                Assertions.fail("Request '" + request.getUrl() + "' is not set up");
-                return Mono.empty();
+        if (query == null) {
```
doncem: Shall we wait for other PRs to get in? all this establishment was under construction and product of all investigations that took course. we can review later

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
doncem: then it's not a processor. it's just a service which cleans up containers 😉

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
doncem: unless it's a service class so some parts are ok (like method construct), some parts not (like logging to begin with)

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleaner.java
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+        this.storageClient = storageClient;
+        this.envelopeRepository = envelopeRepository;
+    }
+
+    public void process(String containerName) {
+        logger.info("Started deleting dispatched blobs from container {}", containerName);
+
+        BlobContainerClient containerClient = storageClient.getBlobContainerClient(containerName);
+
```
doncem: no need to get clients if there is nothing to delete

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,175 @@
...
+    private BlobClient blobClient3;
+
+    @Mock
+    private BlobClient blobClient4;
+
+    private static final UUID UUID_1 = UUID.randomUUID();
+    private static final UUID UUID_2 = UUID.randomUUID();
+
+    private static final String FILE_NAME_1 = "file1.zip";
+    private static final String FILE_NAME_2 = "file2.zip";
```
rkondratowicz: Those constants can be removed.
We can refer to fields from `ENVELOPE_1` and `ENVELOPE_2` instead.
It will also clarify what those are in my opinion.

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,175 @@
...
+    private BlobClient blobClient1;
+
+    @Mock
+    private BlobClient blobClient2;
+
+    @Mock
+    private BlobClient blobClient3;
+
+    @Mock
+    private BlobClient blobClient4;
```
rkondratowicz: why so many mocks?

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
rkondratowicz: 🎉

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
rkondratowicz: 3 and 4 are never used

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,175 @@
...
+class ContainerCleanerTest {
+
+    private static final String CONTAINER_NAME = "bulkscan";
+
+    private ContainerCleaner containerCleaner;
+
+    @Mock
+    private EnvelopeRepositoryImpl envelopeRepository;
+
+    private BlobServiceClient storageClient = mock(BlobServiceClient.class);
```
rkondratowicz: can annotation be used here as well?

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+        // when
+        containerCleaner.process(CONTAINER_NAME);
+
+        // then
+        verifyNoInteractions(containerClient);
+    }
+
+    @Test
+    void should_find_blobs_and_delete() {
```
rkondratowicz: and update envelopes in DB

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    @Test
+    void should_handle_non_existing_file() {
+        // given
+        given(envelopeRepository.find(DISPATCHED, false))
+            .willReturn(singletonList(
+                ENVELOPE_1
+            ));
+        given(containerClient.getBlobClient(FILE_NAME_1)).willReturn(blobClient1);
```
rkondratowicz: if file doesn't exist, shouldn't this return null?

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
```diff
@@ -0,0 +1,175 @@
...
+        verify(blobClient1).delete();
+        verify(envelopeRepository).find(DISPATCHED, false);
+        verify(envelopeRepository).markAsDeleted(UUID_1);
+        verifyNoMoreInteractions(envelopeRepository);
+    }
+
+    @Test
+    void should_handle_server_error() {
+        // given
+        // given
```
rkondratowicz: duplication

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
apereverzin: Fixed

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerCleanerTest.java
apereverzin: File names constants are used

PR #85, opened 2020-01-13
Implement envelopes repository

rkondratowicz:


src/main/resources/application.yaml
```diff
@@ -12,6 +12,12 @@ management:
 spring:
   application:
     name: Blob Router Service
+  datasource:
+    url: jdbc:postgresql://${DB_HOST}:${DB_PORT}/${DB_NAME}${DRAFT_STORE_DB_CONN_OPTIONS:}
```
nitinprabhu: DRAFT_STORE_DB_CONN_OPTIONS ?

rkondratowicz:
@nitinprabhu I've added chart changes to this PR so that we can spot if it's correct

charts/reform-scan-blob-router/values.yaml
```diff
@@ -19,6 +19,12 @@ java:
         - crime-storage-account-primary-key
         - flyway-password
   environment:
+    DB_HOST: blob_router-{{ .Values.global.environment }}.postgres.database.azure.com
+    DB_PORT:'5432'
+    DB_NAME: blob_router
+    DB_USER: blob_router@blob_router-{{ .Values.global.environment }}
+    DRAFT_STORE_DB_CONN_OPTIONS: ?sslmode=require
```
nitinprabhu: draft store ?

src/main/resources/application.yaml
```diff
@@ -12,6 +12,12 @@ management:
 spring:
   application:
     name: Blob Router Service
+  datasource:
+    url: jdbc:postgresql://${DB_HOST}:${DB_PORT}/${DB_NAME}${DB_CONN_OPTIONS:}
```
nitinprabhu: do we nee empty value if DB_CONN_OPTIONS is not set. We don't need : ?

src/main/resources/application.yaml
rkondratowicz: oops

src/main/resources/application.yaml
rkondratowicz: removed

charts/reform-scan-blob-router/values.yaml
```diff
@@ -19,6 +19,12 @@ java:
         - crime-storage-account-primary-key
         - flyway-password
   environment:
+    DB_HOST: blob_router-{{ .Values.global.environment }}.postgres.database.azure.com
+    DB_PORT: '5432'
+    DB_NAME: blob_router
+    DB_USER: blob_router@blob_router-{{ .Values.global.environment }}
+    DB_CONN_OPTIONS: ?sslmode=require
+
```
aliveni: does this change need to bump the chart version?

rkondratowicz: oops, I've updated it, but then merged new master with bumped version. Need to bump again

rkondratowicz: done

PR #87, opened 2020-01-13
Add method for finding envelope by filename and container

rkondratowicz:
As requested by Leszek.
This will be used when processing blobs to check whether file should be skipped or not.

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryImplTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryImplTest.java
@@ -135,6 +135,37 @@ public class EnvelopeRepositoryImplTest {
         });
     }
 
+    @Test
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryImplTest.java
```diff
@@ -135,6 +135,37 @@ void should_find_deleted_envelopes_by_status() {
...
+
+        // then
+        assertThat(result).hasValueSatisfying(envelope -> {
+            assertThat(envelope.fileName).isEqualTo(fileName);
+            assertThat(envelope.container).isEqualTo(container);
+        });
+    }
+
+    @Test
+    void should_return_empty_optional_when_envelope_for_given_container_and_file_name_does_not_exist() {
```
lgonczar: Would be perfect if we checked that we get None when file name matches but container doesn't and when container matches but file name doesn't.

PR #88, opened 2020-01-13
Fix db url for flyway

rkondratowicz:


rkondratowicz:
flux version bump: https://github.com/hmcts/cnp-flux-config/pull/1812

PR #89, opened 2020-01-13
Add blob processor

doncem:

### Change description ###

Add blob processor which

- leases blob
- dispatches it to configured storage
- releases the lease

Depends on #79 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
@@ -1,0 +1,45 @@
+package uk.gov.hmcts.reform.blobrouter.tasks.processors;
+
+import com.azure.core.test.TestBase;
+import com.azure.storage.blob.BlobServiceClient;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import uk.gov.hmcts.reform.blobrouter.services.storage.BlobDispatcher;
+import uk.gov.hmcts.reform.blobrouter.util.StorageClientsHelper;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+class BlobProcessorTest extends TestBase {
+
+    private static final String NEW_BLOB_NAME = "new.blob";
+    private static final String CONTAINER = "bulkscan";
+    private static final String BOGUS_CONTAINER = "bogus";
+
+    private BlobProcessor blobProcessor;
+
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static final String NEW_BLOB_NAME = "new.blob";
+    private static final String CONTAINER = "bulkscan";
+    private static final String BOGUS_CONTAINER = "bogus";
+
+    private BlobProcessor blobProcessor;
+
+    @BeforeAll
+    static void setUpTestMode() {
+        StorageClientsHelper.setAzureTestMode();
```
nitinprabhu: it would be nice if you could explain what this does ?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+
+        try {
+            BlobClient blobClient = storageClient
+                .getBlobContainerClient(containerName)
+                .getBlobClient(blobName);
+            BlobLeaseClient leaseClient = new BlobLeaseClientBuilder()
+                .blobClient(blobClient)
+                .buildClient();
+
+            leaseClient.acquireLease(60);
```
nitinprabhu: we should acquire lease only after we have checked blob doesn't exist in database as otherwise it would be difficult for delete job to acquire lease

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+            byte[] rawBlob = tryToDownloadBlob(blobClient);
+
+            dispatcher.dispatch(blobName, rawBlob, containerName);
+
+            leaseClient.releaseLease();
```
nitinprabhu: probably best this goes in finally ?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
doncem: hm. can add db dependency then on next pr. either-way, lease will expire in 60s. and with finally block mentioned below - immediately :)

PR #91, opened 2020-01-14
Add storage account client beans for Bulk Scan and Crime

lgonczar:

### Change description ###

Add storage account client beans for Bulk Scan and Crime.

Relies on https://github.com/hmcts/blob-router-service/pull/90.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
@@ -1,5 +1,7 @@
 package uk.gov.hmcts.reform.blobrouter.config;
 
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
 import com.azure.storage.common.StorageSharedKeyCredential;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
```diff
@@ -15,4 +17,18 @@ public StorageSharedKeyCredential getStorageSharedKeyCredential(
     ) {
         return new StorageSharedKeyCredential(accountName, accountKey);
     }
+
+    @Bean("bulkscan-storage-client")
+    public static BlobServiceClient getBulkScanStorageClient(
+        @Value("storage.bulkscan.connection-string") String connectionString
```
nitinprabhu: do we use connection string or storage key and account name for bulk scan storage ?
 or have decided to use connection string for both ?

lgonczar: Oh, I forgot to mention. This PR is based on https://github.com/hmcts/blob-router-service/pull/90, where I'm replacing account name + key with connection string for crime, and for consistency also for bulk scan.

PR #93, opened 2020-01-14
Fix db conn string

rkondratowicz:


rkondratowicz:
https://github.com/hmcts/cnp-flux-config/pull/1819

PR #99, opened 2020-01-15
Extend configuration to support Crime

lgonczar:

### Change description ###

Extend configuration to support Crime

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/Chart.yaml
+++ b/charts/reform-scan-blob-router/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for Blob Router Service
 name: reform-scan-blob-router
 home: https://github.com/hmcts/blob-router-service
-version: 0.0.13
+version: 0.0.14
 maintainers:
   - name: HMCTS BSP Team
     email: bspteam@hmcts.net
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -27,7 +27,9 @@ java:
     FLYWAY_URL: jdbc:postgresql://reform-scan-blob-router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/blob_router?sslmode=require
     FLYWAY_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
     FLYWAY_SKIP_MIGRATIONS: true
+
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 30
+    CRIME_DESTINATION_CONTAINER: bs-sit-scans-received
   image: 'hmctspublic.azurecr.io/reform-scan/blob-router:latest'
 reformblobstorage:
   enabled: false
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/config/BlobDestination.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/config/BlobDestination.java
@@ -1,0 +1,6 @@
+package uk.gov.hmcts.reform.blobrouter.config;
+
+public enum BlobDestination {
+    BULKSCAN,
+    CRIME
+}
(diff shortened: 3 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/config/BlobDestination.java
```diff
@@ -0,0 +1,6 @@
+package uk.gov.hmcts.reform.blobrouter.config;
+
+public enum BlobDestination {
+    BULKSCAN,
```
nitinprabhu: do you meant CFT here instead of bulkscan ?

charts/reform-scan-blob-router/values.yaml
```diff
@@ -27,7 +27,9 @@ java:
     FLYWAY_URL: jdbc:postgresql://blob_router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/blob_router?sslmode=require
     FLYWAY_USER: blob_router@blob_router-{{ .Values.global.environment }}
     FLYWAY_SKIP_MIGRATIONS: true
+
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 30
+    CRIME_DESTINATION_CONTAINER: bs-sit-scans-received
```
nitinprabhu: needs chart version bumping.

src/main/resources/application.yaml
```diff
@@ -23,19 +23,26 @@ service:
   storage-config:
     - name: bulkscan
       sas-validity: 300 #In seconds
+      blob-destination: bulkscan
```
nitinprabhu: not sure blob-destination would be a proper name as it would be mean blob is getting copied to bulkscan but actually it would be cmc or probate and so on. We can rename it later not required for this PR

src/main/java/uk/gov/hmcts/reform/blobrouter/config/BlobDestination.java
lgonczar: I originally called it CFT, but then realised the terminology we've been using in this repository is bulkscan. Don't mind changing it, but we'll lose some consistency.

rkondratowicz: worth adding a javadoc comment on this class imho.

lgonczar: Will do. Maybe we can decide what convention we want to go for here?

lgonczar: Plus, any ideas for name to replace `blob-destination` with?

rkondratowicz: perhaps `target-storage-account`?

rkondratowicz: and value `bulkscan` (for the storage account) is clearer to me

src/main/resources/application.yaml
rkondratowicz: how about `target-storage-account`? Or  `storage-account-desination`?

lgonczar:
Could you have a look at the changed names, please?

nitinprabhu:
> Could you have a look at the changed names, please?

looks good to me. Chart version needs bumping

charts/reform-scan-blob-router/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for Blob Router Service
 name: reform-scan-blob-router
 home: https://github.com/hmcts/blob-router-service
-version: 0.0.13
+version: 0.0.14
```
nitinprabhu: as usual flux would also need an update

PR #100, opened 2020-01-15
Add repository to BlobProcessor

doncem:

### Change description ###

Adding repository to the processor. Only processing blob if it is not yet registered in the db. Creating the record afterwords

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
@@ -4,17 +4,40 @@ import com.azure.core.test.TestBase;
 import com.azure.storage.blob.BlobServiceClient;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import uk.gov.hmcts.reform.blobrouter.data.DbHelper;
+import uk.gov.hmcts.reform.blobrouter.data.EnvelopeRepository;
+import uk.gov.hmcts.reform.blobrouter.data.model.NewEnvelope;
+import uk.gov.hmcts.reform.blobrouter.data.model.Status;
 import uk.gov.hmcts.reform.blobrouter.services.storage.BlobDispatcher;
 import uk.gov.hmcts.reform.blobrouter.util.StorageClientsHelper;
 
-import static org.assertj.core.api.Assertions.assertThatCode;
+import java.util.UUID;
 
+import static java.time.Instant.now;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+@ActiveProfiles("db-test")
+@SpringBootTest
 class BlobProcessorTest extends TestBase {
 
     private static final String NEW_BLOB_NAME = "new.blob";
     private static final String CONTAINER = "bulkscan";
     private static final String BOGUS_CONTAINER = "bogus";
 
+    @Autowired
+    private EnvelopeRepository envelopeRepository;
+    @Autowired
+    private DbHelper dbHelper;
+
+    private BlobDispatcher dispatcher;
     private BlobProcessor blobProcessor;
 
     @BeforeAll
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -19,16 +24,32 @@
...
     ) {
         this.storageClient = storageClient;
         this.dispatcher = dispatcher;
+        this.envelopeRepository = envelopeRepository;
     }
 
     public void process(String blobName, String containerName) {
+        envelopeRepository.find(blobName, containerName).ifPresentOrElse(
+            envelope -> logger.warn(
+                "Envelope already processed in system. ID: {}, filename: {}, container: {}, state: {}",
```
rkondratowicz: worth adding info on what we do about. ie, skipping processing.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -26,20 +49,52 @@ static void setUpTestMode() {
...
+    void should_not_process_blob_if_present_in_db() {
+        // given
+        NewEnvelope newEnvelope = new NewEnvelope(
+            CONTAINER,
+            NEW_BLOB_NAME,
+            now(),
+            now().plusSeconds(100),
+            Status.REJECTED
+        );
+        UUID id = envelopeRepository.insert(newEnvelope);
```
rkondratowicz: `id ` not used

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -45,16 +66,26 @@ public void process(String blobName, String containerName) {
             byte[] rawBlob = tryToDownloadBlob(blobClient);
 
             dispatcher.dispatch(blobName, rawBlob, containerName);
+            UUID envelopeId = envelopeRepository.insert(createNewEnvelope(
```
lgonczar: We should first check if the envelope already exists. Only if it doesn't, we should acquire lease on the file and carry on. Otherwise, we'd be flooded by "lease already acquired" cases, which we can avoid.

rkondratowicz: it's done above

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -45,16 +66,26 @@ public void process(String blobName, String containerName) {
...
 
-            logger.info("Finished processing {} from {} container", blobName, containerName);
+            logger.info(
+                "Finished processing {} from {} container. New envelope ID: {}",
+                blobName,
+                containerName,
+                envelopeId
+            );
         } catch (Exception exception) {
             logger.error("Error occurred while processing {} from {}", blobName, containerName, exception);
```
lgonczar: If we failed to acquire lease, we should just log it as info and continue with other blobs.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -19,16 +24,32 @@
...
+        EnvelopeRepository envelopeRepository
     ) {
         this.storageClient = storageClient;
         this.dispatcher = dispatcher;
+        this.envelopeRepository = envelopeRepository;
     }
 
     public void process(String blobName, String containerName) {
+        envelopeRepository.find(blobName, containerName).ifPresentOrElse(
+            envelope -> logger.warn(
```
rkondratowicz: did you consider formatting it like this:
```java
envelopeRepository
  .find(blob, containerName)
  .ifPresentOrElse(
    envelope -> ...
    () -> ...
  )
)
```
?

Up to you.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
lgonczar: aaa, ok, my bad

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
lgonczar: Not even sure if it should be a warning - this will happen from time to time, because deletion job will be a bit behind.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
doncem: After all this usage we actually will only get (as far as I have expected) `BlobStorageException` so we need to distinguish between different steps. Perhaps need to extract each step to isolated/different method call so to be able to catch all exceptions individually

rkondratowicz: Good point problem acquiring lease.
Acquiring lease and handling exception existed before this PR, so I think it would be better to address this in a separate PR.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
doncem: i would vote for debug, but please propose what we would want?

rkondratowicz: info

PR #107, opened 2020-01-15
Config for blob deletion

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-970

### Change description ###
Defined config for scheduled blob deletion task

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -28,7 +28,6 @@ java:
     FLYWAY_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
     FLYWAY_SKIP_MIGRATIONS: true
 
-    DELETE_DISPATCHED_FILES_ENABLED: "true"
     DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
 
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 30
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -28,6 +28,9 @@ java:
     FLYWAY_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
     FLYWAY_SKIP_MIGRATIONS: true
 
+    DELETE_DISPATCHED_FILES_ENABLED: "true"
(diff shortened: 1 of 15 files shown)


rkondratowicz:
Please add usage

infrastructure/aat.tfvars
```diff
@@ -1 +1,3 @@
 api_test_valid_certificate_thumbprint = "A612B39D42D0D72DEB738F4DF77EF8997023E078"
+
+delete_dispatched_files_cron = "0 0/10 * * * *"
```
lgonczar: This should be something more frequent, every ten seconds for example, so that functional tests don't wait too long for files to be deleted.

src/main/java/uk/gov/hmcts/reform/blobrouter/util/TimeZones.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.hmcts.reform.blobrouter.util;
+
+public final class TimeZones {
```
lgonczar: We should probably get rid of this file from this PR

charts/reform-scan-blob-router/values.yaml
```diff
@@ -28,6 +28,8 @@ java:
     FLYWAY_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
     FLYWAY_SKIP_MIGRATIONS: true
 
+    DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
```
rkondratowicz: looks extremely rare, considering the main job will run every ~30sec.

lgonczar: In test envs it runs every 10 seconds. In production, what's the point of making it run any more frequently than once in 10 minutes?

lgonczar: We could even, perhaps, delete files at the end of the day. This way we would be able to resend the files in case our client loses them.

lgonczar: Plus, the main job shouldn't run every 30 seconds, but approx. 10 minutes, as well. Otherwise, each file would be checked for readiness by each instance approx. 60 times.

apereverzin: I agree with Leszek - no point to run delete job as frequently as processing job

lgonczar: Worth discussing within the team

PR #108, opened 2020-01-15
Check if blob is ready for processing

rkondratowicz:
Using checker class from previous PR in the blob processing job.

Best to review with whitespace changes **OFF**.

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
@@ -6,11 +6,13 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.context.ActiveProfiles;
 import uk.gov.hmcts.reform.blobrouter.data.DbHelper;
 import uk.gov.hmcts.reform.blobrouter.data.EnvelopeRepository;
 import uk.gov.hmcts.reform.blobrouter.data.model.NewEnvelope;
 import uk.gov.hmcts.reform.blobrouter.data.model.Status;
+import uk.gov.hmcts.reform.blobrouter.services.BlobReadinessChecker;
 import uk.gov.hmcts.reform.blobrouter.services.storage.BlobDispatcher;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Optional;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(MockitoExtension.class)
+class BlobProcessorTest {
```
rkondratowicz: note that unit tests for `BlobProcessor` did not exist before, but adding any other relevant tests for this class is not in the scope of this PR.

Only doing checks around blob readiness.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -56,7 +62,15 @@ protected void beforeTest() {
...
-        blobProcessor = new BlobProcessor(storageClient, dispatcher, envelopeRepository, leaseClientProvider);
+        when(readinessChecker.isReady(any())).thenReturn(true); // disable processing delay.
+
+        blobProcessor = new BlobProcessor(
+            storageClient,
+            dispatcher,
+            readinessChecker,
+            envelopeRepository,
+            leaseClientProvider
+        );
```
rkondratowicz: line got too long with the new parameter, sorry.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -64,24 +68,32 @@ private void processBlob(String blobName, String containerName) {
...
-                blobClient.getProperties().getCreationTime().toInstant()
-            ));
-
-            logger.info(
-                "Finished processing {} from {} container. New envelope ID: {}",
-                blobName,
-                containerName,
-                envelopeId
-            );
+            if (this.readinessChecker.isReady(blobClient.getProperties().getCreationTime().toInstant())) {
```
doncem: We could assign creation time to local variable as it is used twice in the same scope. Will also make if statement a little bit more sensible

rkondratowicz: good point

PR #111, opened 2020-01-16
Enable blob scan task

doncem:

### Change description ###

Includes all annotations to make blob task working:

- schedule (enabling is already done by introduction of shedlock library/configuration in previous prs)
- components/beans
- bean qualifiers as multiple clients are in play

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
@@ -1,5 +1,7 @@
 package uk.gov.hmcts.reform.blobrouter.config;
 
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
 import com.azure.storage.blob.specialized.BlobLeaseClientBuilder;
 import com.azure.storage.common.StorageSharedKeyCredential;
 import org.springframework.beans.factory.annotation.Value;
@@ -16,6 +18,16 @@ public class StorageConfiguration {
         @Value("${storage.account-key}") String accountKey
     ) {
         return new StorageSharedKeyCredential(accountName, accountKey);
+    }
+
+    // not used as needs test context so it can actually be build
+    @Bean({
+        "storage-client",
+        "bulkscan-storage-client",
+        "crime-storage-client"
+    })
+    public BlobServiceClient getStorageClient(StorageSharedKeyCredential credentials) {
+        return new BlobServiceClientBuilder().credential(credentials).buildClient();
     }
 
     @Bean
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
@@ -27,6 +27,11 @@ public class StorageConfiguration {
         return blobClient -> new BlobLeaseClientBuilder().blobClient(blobClient).buildClient();
     }
 
+    @Bean("storage-client")
+    public BlobServiceClient getStorageClient(StorageSharedKeyCredential credentials) {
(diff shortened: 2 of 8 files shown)


src/main/resources/application.yaml
```diff
@@ -55,3 +55,5 @@ scheduling:
     delete-dispatched-files:
       enabled: true
       cron: ${DELETE_DISPATCHED_FILES_CRON}
+    scan:
+      delay: 30000 # In milliseconds
```
nitinprabhu: do we need 30 secs delay by default. Better to configure this per environment where AAT/Preview can have minimum delay ?

src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
```diff
@@ -27,6 +27,11 @@ public LeaseClientProvider getLeaseClientProvider() {
         return blobClient -> new BlobLeaseClientBuilder().blobClient(blobClient).buildClient();
     }
 
+    @Bean("storage-client")
+    public BlobServiceClient getStorageClient(StorageSharedKeyCredential credentials) {
```
nitinprabhu: there are 3 storage clients now ?

doncem: yes. `storage-client` which is service's own one. and 2 prefixed ones noting what are they

src/main/resources/application.yaml
doncem: there was no need to customise this property. at the time of creating this pr functional test configuration was still underway and this property can be simply included into `values.aat` and `values.preview` without amending template which would need chart version upgrade

rkondratowicz: we can at least add env var here so that (clear) overwriting is possible

nitinprabhu: It needs an env var else you will have to bootstrap from vault which is not required for this property.

src/main/resources/application.yaml
```diff
@@ -55,3 +55,5 @@ scheduling:
     delete-dispatched-files:
       enabled: true
       cron: ${DELETE_DISPATCHED_FILES_CRON}
+    scan:
+      delay: ${TASK_SCAN_DELAY:30000} # In milliseconds
```
nitinprabhu: This PR won't work unless we have this in our charts. I assume you are creating a separate PR ?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
```diff
@@ -18,6 +20,16 @@ public StorageSharedKeyCredential getStorageSharedKeyCredential(
...
 
+    // not used as needs test context so it can actually be build
+    @Bean({
+        "storage-client",
+        "bulkscan-storage-client",
+        "crime-storage-client"
+    })
+    public BlobServiceClient getStorageClient(StorageSharedKeyCredential credentials) {
+        return new BlobServiceClientBuilder().credential(credentials).buildClient();
+    }
```
rkondratowicz: if those beans are not used, do they need to be defined? (and point to the same container)

doncem: Need to be defined to load application context for certain integration tests. Otherwise:

> org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.azure.storage.blob.BlobServiceClient' available

This was achieved by just returning a `null` here as well

lgonczar: right, it would probably be best to limit the beans to the minimum that the tests need

src/main/resources/application.yaml
nitinprabhu: sorry missed that it has default. As discussed let's remove the default one once we have the chart PR

PR #114, opened 2020-01-16
Move default to chart config

doncem:

### Change description ###

Using defaults is not recommended and env var was dedicated to be used and configured in chart after discussion during #111 

Flux config PR is on the way

```
[ ] Yes
[x] No
```


doncem:
I did not want to go so low as 4000 so making 1000 in values.yaml would trigger an eye twitch for me 😂

PR #116, opened 2020-01-16
Add environment variables for storage signature

aliveni:

### Change description ###
Add environment variables for storage signature algorithm and public key.
Flux config Pr  - https://github.com/hmcts/cnp-flux-config/pull/1844

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.yaml
```diff
@@ -32,6 +32,8 @@ java:
     TASK_SCAN_DELAY: "4000" # in millis
 
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 30
+    STORAGE_BLOB_SIGNATURE_ALGORITHM: "sha256withrsa"
+    STORAGE_BLOB_PUBLIC_KEY: "nonprod_public_key.der"
```
lgonczar: How is it going to work? Won't we have the public key in memory somewhere?

aliveni: I added it in #117

src/main/resources/application.yaml
```diff
@@ -48,6 +48,8 @@ flyway:
   skip-migrations: ${FLYWAY_SKIP_MIGRATIONS}
 
 storage-blob-processing-delay-in-minutes: ${STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES}
+storage-signature-algorithm: ${STORAGE_BLOB_SIGNATURE_ALGORITHM:sha256withrsa}  # none or sha256withrsa
+public_key_der_file: ${STORAGE_BLOB_PUBLIC_KEY:test_public_key.der} # public key file in der format
```
rkondratowicz: we probably want to have no default here just in case

charts/reform-scan-blob-router/values.yaml
mdayican: do we need  special config for Preview?

aliveni:
> Need chart version bumping ?

Yes, it's outdated. bumped the version again

PR #120, opened 2020-01-16
Add url for main storage client to work

doncem:

### Change description ###

Setting endpoint to storage client. This is mandatory as library doesn't build from from credentials which has all properties needed to build endpoint. Right now we end up with this error:

> java.net.MalformedURLException: no protocol: /bulkscan no protocol: /bulkscan

when trying to get container client

Value is set during this PR: #123 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
@@ -28,8 +28,14 @@ public class StorageConfiguration {
     }
 
     @Bean("storage-client")
-    public BlobServiceClient getStorageClient(StorageSharedKeyCredential credentials) {
-        return new BlobServiceClientBuilder().credential(credentials).buildClient();
+    public BlobServiceClient getStorageClient(
+        StorageSharedKeyCredential credentials,
+        @Value("${storage.url}") String storageUrl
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
```diff
@@ -29,7 +29,10 @@ public LeaseClientProvider getLeaseClientProvider() {
 
     @Bean("storage-client")
     public BlobServiceClient getStorageClient(StorageSharedKeyCredential credentials) {
-        return new BlobServiceClientBuilder().credential(credentials).buildClient();
+        return new BlobServiceClientBuilder()
+            .credential(credentials)
+            .endpoint(String.format("https://%s.blob.core.windows.net", credentials.getAccountName()))
```
lgonczar: It would be perfect if this URL came from configuration. This is already done this way for functional tests.

nitinprabhu: I think windows.net won't work even with blob router storage account.

doncem: yep. need to add app gw

PR #119, opened 2020-01-16
add bulk-scan-processor client to get sas token

mdayican:
### Change description ###

add bulk-scan-processor client to get sas token

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #121, opened 2020-01-17
Proposal: retrieve blob storage clients dynamically

lgonczar:
### Change description ###

This PR will not go to master - it's just a proposal to be verified by the team.

Background: crime storage client doesn't change - we can use the same one throughout the whole life of the application. Bulk scan storage client, however, is built based on SAS tokens, which expire every few hours. Therefore, bulk scan clients need to be provided dynamically and crime needs to fit in the existing solution.

I've based this PR on https://github.com/hmcts/blob-router-service/pull/119, which contains SAS token endpoint client.

```
[ ] Yes
[x] No
```


rkondratowicz:
looks good

PR #126, opened 2020-01-17
Fix configuration for accessing Bulk Scan blob storage

lgonczar:

### Change description ###

Provide the right configuration values for the application and its tests, necessary for uploading blobs to bulk scan blob storage.

This PR will be followed by one that prepares dynamic SAS token-based clients.

```
[ ] Yes
[x] No
```


Jenkinsfile_CNP
```diff
@@ -91,10 +92,8 @@ withPipeline(type, product, component) {
       env.TEST_STORAGE_ACCOUNT_URL = "https://${env.TEST_STORAGE_ACCOUNT_NAME}.blob.core.windows.net"
       env.TEST_STORAGE_ACCOUNT_KEY = previewAccountKey(kubectl, storageSecret, namespace)
 
-      env.STORAGE_BULKSCAN_CONNECTION_STRING = storageAccountConnectionString(
-        previewAccountName(kubectl, bulkScanStorageAccountSecret, namespace),
-        previewAccountKey(kubectl, bulkScanStorageAccountSecret, namespace)
-      )
+      def bulkScanStorageAccountName = previewAccountName(kubectl, bulkScanStorageAccountSecret, namespace)
+      env.STORAGE_BULKSCAN_URL = "https://${bulkScanStorageAccountName}.blob.core.windows.net"
```
nitinprabhu: for preview there is no app gw as we use osba so existing config should work ?

lgonczar: but do we want different behaviour for preview?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
```diff
@@ -17,9 +17,7 @@
 
     private final BlobServiceClient bulkscanStorageClient;
 
-    public BlobDispatcher(
-        @Qualifier("bulkscan-storage-client") BlobServiceClient bulkscanStorageClient
-    ) {
+    public BlobDispatcher(BlobServiceClient bulkscanStorageClient) {
```
lgonczar: In the next PR BlobDispatcher is going to have client provide injected. I've removed the qualifier, because that client doesn't exist anymore. It wasn't a working client, anyway.

Jenkinsfile_CNP
nitinprabhu: it has to be for now as we don't have app gw for preview (this is what we have in processor)

lgonczar: This is the storage account's URL

lgonczar: needs discussion

nitinprabhu: apologies this is correct

PR #131, opened 2020-01-17
Provide blob storage clients dynamically

lgonczar:

### Change description ###

This PR covers the following:
- Provide blob storage clients dynamically to blob dispatcher
- Configuration changes to enable using bulk scan processor blob storage (changes from #130)

These changes are very much related and it's hard to introduce them one after another, as the configuration changes would break the application.

This change will be followed by:
- adjusting the solution to use crime for `crime` container
- functional tests for CFT

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/controllers/SasTokenControllerExceptionTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/controllers/SasTokenControllerExceptionTest.java
@@ -3,10 +3,12 @@ package uk.gov.hmcts.reform.blobrouter.controllers;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.web.servlet.MockMvc;
 import uk.gov.hmcts.reform.blobrouter.exceptions.UnableToGenerateSasTokenException;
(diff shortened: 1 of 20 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/controllers/SasTokenControllerExceptionTest.java
```diff
@@ -24,6 +26,9 @@
     @Autowired
     private MockMvc mockMvc;
 
+    @MockBean
+    private BlobServiceClientProvider blobServiceClientProvider;
+
```
lgonczar: Introduction of this service, which uses bulk scan processor client, caused FeignContext-related problem that I couldn't solve quickly. I'd be grateful if somebody has a better solution.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobServiceClientProvider.java
```diff
@@ -0,0 +1,44 @@
...
+            case BULKSCAN:
+                // retrieving a SAS token every time we're getting a client, but this will be cached in the future
+                return new BlobServiceClientBuilder()
+                    .sasToken(bulkScanSasTokenClient.getSasToken(containerName).sasToken)
+                    .endpoint(bulkScanStorageUrl)
+                    .buildClient();
+            case CRIME:
+                return crimeClient;
+            default:
+                throw new UnknownStorageAccountException(
```
nitinprabhu: this scenario is not covered by tests but should be fine to be done later as a separate Pr

lgonczar: It's impossible to pass an unrecognised account value. That would only happen if someone added a new value to the enum but forgot to handle it in this class.

PR #135, opened 2020-01-17
Stop BlobDispatcher from catching exceptions

lgonczar:
### Change description ###

Stop BlobDispatcher from catching exceptions, so that BlobProcessor doesn't save envelopes when upload failed.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
@@ -28,30 +28,20 @@ public class BlobDispatcher {
     ) {
         logger.info("Uploading {} to {} container. Storage: {}", blobName, destinationContainer, targetStorageAccount);
 
-        try {
-            getContainerClient(targetStorageAccount, destinationContainer)
-                .getBlobClient(blobName)
-                .getBlockBlobClient()
-                .upload(
-                    new ByteArrayInputStream(blobContents),
-                    blobContents.length
-                );
+        getContainerClient(targetStorageAccount, destinationContainer)
+            .getBlobClient(blobName)
+            .getBlockBlobClient()
+            .upload(
+                new ByteArrayInputStream(blobContents),
+                blobContents.length
+            );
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
```diff
@@ -28,30 +28,20 @@ public void dispatch(
...
-                destinationContainer,
-                targetStorageAccount
-            );
-        } catch (Exception exception) {
-            logger.error(
-                "Error occurred while uploading {} to {} container. Storage: {}",
-                blobName,
-                destinationContainer,
-                exception,
-                targetStorageAccount
```
lgonczar: The error is logged by BlobProcessor

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -67,6 +68,23 @@ void should_not_process_file_if_it_is_not_ready_yet() {
...
+
+        willThrow(new RuntimeException("Test exception"))
+            .given(blobDispatcher)
+            .dispatch(any(), any(), any(), any());
+
+        // when
+        blobProcessor.process("envelope.zip", "container1");
+
+        // then
+        verify(envelopeRepo, never()).insert(any());
```
rkondratowicz: may be worth verifying that method throwing exception was called

lgonczar: Good idea, done

lgonczar:
> we could throw specific exception so that exception handler (injected/implemented upstream) could handle each specific exception individually:
> 
> * send notification
> * reject
> * leave for subsequent processing

BlobProcessor will probably end up catching at least two kinds of exceptions - one for when we need to reject an envelope. A failure to upload a blob doesn't require any special action, though - it can be handled as any other processing failures.

PR #151, opened 2020-01-21
Add envelope controller

rkondratowicz:
It will be used in functional tests to verify that envelope has been dispatched.

May also come in handy on any environment when checking the status of given file.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeController.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeController.java
@@ -8,6 +8,7 @@ import org.springframework.web.bind.annotation.RestController;
 import uk.gov.hmcts.reform.blobrouter.data.EnvelopeRepository;
 import uk.gov.hmcts.reform.blobrouter.data.model.Envelope;
 import uk.gov.hmcts.reform.blobrouter.exceptions.EnvelopeNotFoundException;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeController.java
```diff
@@ -0,0 +1,31 @@
...
+    private final EnvelopeRepository envelopesRepo;
+
+    public EnvelopeController(EnvelopeRepository envelopesRepo) {
+        this.envelopesRepo = envelopesRepo;
+    }
+
+    @GetMapping()
+    public Envelope findEnvelope(
+        @RequestParam("file_name") String fileName,
+        @RequestParam("container") String container
```
lgonczar: Why not path params, i.e. /containers/{}/envelopes/{}? We can represent these envelopes are resources.

rkondratowicz: it would work if we were to send envelope ID, but otherwise not sure

rkondratowicz: we could later on expand this endpoint to take more parameters and becore a general 'search' endpoint

rkondratowicz:
> * Does `Envelope` have to have Jackson annotations?
> * Do we need to create separate model for responses?

Added new model (with jackson annotations)

PR #159, opened 2020-01-22
Process containers in random order

rkondratowicz:
To make it less probable for 2 instances of processor competing for the same files.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
@@ -8,8 +8,11 @@ import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.blobrouter.config.ServiceConfiguration;
 import uk.gov.hmcts.reform.blobrouter.tasks.processors.ContainerProcessor;
 
+import java.util.Collections;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
```diff
@@ -36,17 +39,20 @@ public BlobDispatcherTask(
     public void run() {
         logger.info("Started {} job", TASK_NAME);
 
-        getAvailableContainers().forEach(containerProcessor::process);
+        List<String> containers = getAvailableContainers();
+        Collections.shuffle(containers);
```
rkondratowicz: I wish this returned a value... 😢

lgonczar:
Containers are fine and this PR can be merged. In a typical scenario, however, this shuffling will make little difference, as all jobs will quickly end up in the same container - the one where files are.

We can shuffle blobs in a future PR. `PageIterable` is easily transformable to List.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
doncem: We could try use something different this time - avoiding change in public method can just collect in private using `Collectors.collectingAndThen()`. Just a random idea

PR #161, opened 2020-01-23
Add bulkscan functional test

doncem:

### Change description ###

Awaiting until blob is deleted and verifying db values.

Since reusing same source created for crime and service doesn't care about it's contents - removing the prefix of 'crime' and reusing same zips for different (bulkscan) container

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/BulkScanBlobDispatchingTest.java
```diff
@@ -0,0 +1,57 @@
...
+        // upload bulkscan file with unique name
+        String fileName = randomFileName();
+
+        byte[] internalZipContent = toByteArray(
+            getResource("test-data/envelope/envelope.zip")
+        );
+
+        byte[] wrappingZipContent = createZipArchive(
+            asList("test-data/envelope/envelope.zip", "test-data/envelope/signature")
+        );
```
lgonczar: This is fine, although for CFT we can reduce these two statements to just one. For crime it matters that there's a zip containing another zip, because blob-router is supposed to extract it and only send the internal one. For CFT blobs content doesn't matter much (although in the future it will, to some extent). Up to you.

lgonczar:
We might also need to point the blob router to the processor and its blob storage in aat for preview, otherwise, there will be a problem with SAS token retrieval.

doncem:
> We might also need to point the blob router to the processor and its blob storage in aat for preview, otherwise, there will be a problem with SAS token retrieval.

missed that! had lot's of changes and reverted everything before starting doing this job

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/BulkScanBlobDispatchingTest.java
doncem: non-repudiation will be applied to all blobs disregarding the source container. so even content doesn't matter, but signature does

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/BulkScanBlobDispatchingTest.java
```diff
@@ -0,0 +1,57 @@
...
+            .until(
+                () -> !blobExists(blobRouterStorageClient, CONTAINER, fileName)
+            );
+
+        // and
+        RestAssured
+            .given()
+            .baseUri(config.blobRouterUrl)
+            .relaxedHTTPSValidation()
+            .get("/envelopes?file_name=" + fileName + "&container=" + CONTAINER)
```
rkondratowicz: I think you **can** also set the query params using
```java
.queryParam("file_name", fileName)
.queryParam("container", container)
```
and leave the url as `/envelopes`.

Up to you.

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/BulkScanBlobDispatchingTest.java
lgonczar: fair enough

src/functionalTest/resources/application.conf
```diff
@@ -1,3 +1,4 @@
+source-url=${TEST_URL}
```
rkondratowicz: 'source' seems weird here imho.
Maybe one of the below?
- `app-url`
- `service-url`
- `blob-router-url`

charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -66,6 +68,11 @@ reformblobstorage:
     containers:
       - bulkscan
       - bulkscan-rejected
+      - cmc
+      - divorce
+      - finrem
+      - probate
+      - sscs
```
nitinprabhu: don't need this for preview ?

doncem: The way we access blobs we gather containers from configuration and try to access them all. To avoid unnecessary errors or unknown failures - adding all containers from configuration

PR #168, opened 2020-01-24
Add integration test for blob processor

rkondratowicz:


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor2Test.java
```diff
@@ -0,0 +1,75 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.BDDMockito.given;
+import static uk.gov.hmcts.reform.blobrouter.data.model.Status.DISPATCHED;
+
+@ActiveProfiles("db-test")
+@SpringBootTest
+@ExtendWith(MockitoExtension.class)
+public class BlobProcessor2Test extends BlobStorageBaseTest {
```
rkondratowicz: Putting '2' in the name because `BlobProcessor2Test` already exists and this is a complete rewrite.
I'll remove the original file (and remove this one) in next PR.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -72,7 +72,7 @@ private void processBlob(String blobName, String containerName) {
                 .getBlobContainerClient(containerName)
                 .getBlobClient(blobName);
 
-            Instant blobCreationDate = blobClient.getProperties().getCreationTime().toInstant();
+            Instant blobCreationDate = blobClient.getProperties().getLastModified().toInstant();
```
rkondratowicz: Creation date is null. This might be Azurite bug, not sure.
We use 'last modified' date in bulk-scan-processor as well.

doncem: good catch. good hypothesis too

PR #176, opened 2020-01-28
Use vars

rkondratowicz:
a couple more, following https://github.com/hmcts/blob-router-service/pull/169


rkondratowicz:
> var lover 😍

unnecessary duplication hater ;)

PR #177, opened 2020-01-28
Add repo method for finding envelopes by `status` and `isDeleted`

rkondratowicz:
There's a similar method already that also takes container as parameter, this one doesn't.

It will be used by job moving rejected files to rejected container.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepository.java
@@ -65,6 +65,16 @@ public class EnvelopeRepository {
         );
     }
 
+    public List<Envelope> find(Status status, boolean isDeleted) {
+        return jdbcTemplate.query(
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepository.java
```diff
@@ -65,6 +65,16 @@ public EnvelopeRepository(NamedParameterJdbcTemplate jdbcTemplate, EnvelopeMappe
         );
     }
 
+    public List<Envelope> find(Status status, boolean isDeleted) {
+        return jdbcTemplate.query(
+            "SELECT * FROM envelopes WHERE status = :status AND is_deleted = :isDeleted",
```
doncem: Do we need index for those columns?

nitinprabhu: good point

PR #178, opened 2020-01-28
Remove signature algorithm config

aliveni:

### Change description ###
- Update `ZipVerifiers` class to always verify the signature with `sha256withrsa` algorithm
- Remove signature algorithm config

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.yaml
```diff
@@ -33,7 +33,6 @@ java:
     STORAGE_URL: https://reformscan.{{ .Values.global.environment }}.platform.hmcts.net
     STORAGE_BULKSCAN_URL: https://bulkscan{{ .Values.global.environment }}.blob.core.windows.net
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 30
-    STORAGE_BLOB_SIGNATURE_ALGORITHM: "sha256withrsa"
```
doncem: does that mean we need a separate PR just for config changes supplementing with chart version bump?

(which should then be done after this PR)

src/main/java/uk/gov/hmcts/reform/blobrouter/util/ZipVerifiers.java
```diff
@@ -29,7 +27,7 @@
 import static java.util.Arrays.asList;
 
 /**
- * Signed zip archive verification utilities. Currently 2 modes are supported:
+ * Signed zip archive verification utilities. Currently verifies with sha256withrsa algorithm:
  * <ul>
  * <li>sha256withrsa = sha256 + rsa signature verification</li>
  * </ul>
```
doncem: perhaps no need for ul/li now 😉

charts/reform-scan-blob-router/values.yaml
aliveni: bumped chart version. flux config pr: https://github.com/hmcts/cnp-flux-config/pull/1957

src/main/java/uk/gov/hmcts/reform/blobrouter/util/ZipVerifiers.java
```diff
@@ -29,10 +27,9 @@
...
 
 /**
- * Signed zip archive verification utilities. Currently 2 modes are supported:
- * <ul>
- * <li>sha256withrsa = sha256 + rsa signature verification</li>
- * </ul>
+ * Signed zip archive verification utilities.
+ * Currently verifies with sha256withrsa algorithm:
+ * sha256withrsa = sha256 + rsa signature verification.
  * With the obvious exclusion of the no verification case, a signed zip
```
rkondratowicz: this can be deleted as well

src/main/java/uk/gov/hmcts/reform/blobrouter/util/ZipVerifiers.java
```diff
@@ -112,15 +98,15 @@ static void verifyFileNames(Set<String> fileNames) {
...
+    private static boolean verifySignature(String publicKeyBase64, Map<String, byte[]> entries) {
+        return verifySignature(
             publicKeyBase64,
             entries.get(DOCUMENTS_ZIP),
             entries.get(SIGNATURE_SIG)
         );
     }
 
-    public static void verifySignature(String publicKeyBase64, byte[] data, byte[] signed) {
+    public static boolean verifySignature(String publicKeyBase64, byte[] data, byte[] signed) {
```
rkondratowicz: I'd say it's not related to removing the config entry.
Also, why return boolean if only true is ever returned? (if signature is invalid, in current implementation, an exception is thrown)

aliveni: reverted the change

PR #181, opened 2020-01-29
Add service for moving files to rejected container

rkondratowicz:
It will be called by scheduled job.


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesMover.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.List;
+
+import static java.util.stream.Collectors.groupingBy;
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class RejectedFilesMover {
```
rkondratowicz: I'm open to suggestions on better name :)

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -0,0 +1,104 @@
...
+    public void handle() {
+        List<Envelope> rejectedEnvelopes = envelopeRepository.find(Status.REJECTED, false);
+
+        logger.info("Found {} rejected envelopes", rejectedEnvelopes.size());
+
+        rejectedEnvelopes
+            .stream()
+            .collect(groupingBy(e -> e.container))
+            .keySet()
+            .forEach(container -> {
```
lgonczar: We could just iterate over the grouping, without converting it to a key set. This way we could avoid internal transformation of this collection.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -0,0 +1,104 @@
...
+                            if (!sourceBlob.exists()) {
+                                logger.error(
+                                    "File already deleted. File name: {}. Container: {}",
+                                    envelope.fileName,
+                                    container
+                                );
+                                envelopeRepository.markAsDeleted(envelope.id);
+                            } else {
+                                byte[] blobContent = download(sourceBlob);
+                                upload(targetBlob, blobContent);
```
nitinprabhu: worth logging the facts when file was successfully uploaded

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -0,0 +1,104 @@
...
+        this.envelopeRepository = envelopeRepository;
+    }
+
+    /**
+     * Handles files that were rejected. Ie:
+     * - moves files to container for rejected files
+     * - removes files from the original container
+     * - marks envelopes in the DB as deleted
+     */
+    public void handle() {
```
lgonczar: This method is too long

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private void upload(BlobClient blobClient, byte[] blobContent) {
+        blobClient
+            .getBlockBlobClient()
+            .upload(
+                new ByteArrayInputStream(blobContent),
+                blobContent.length
+            );
+    }
```
lgonczar: We're downloading in order to upload, we could avoid having byte arrays and use steams instead. We can't do this in the main job, as we need to verify the signature and extract the content, but it's possible here.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
rkondratowicz: I want to get container client only once for every container instead of for every file

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
lgonczar: Good point. Both the fact of upload and marking as deleted would be good to know, for troubleshooting.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
rkondratowicz: refactored

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
rkondratowicz: done

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -0,0 +1,103 @@
...
+            .collect(groupingBy(e -> e.container))
+            .forEach((container, envelopes) -> {
+                logger.info("Moving rejected files from container {}", container);
+
+                var sourceContainer = storageClient.getBlobContainerClient(container);
+                var targetContainer = storageClient.getBlobContainerClient(container + REJECTED_CONTAINER_SUFFIX);
+
+                envelopes.forEach(envelope -> handleEnvelope(sourceContainer, targetContainer, envelope));
+            });
+    }
```
lgonczar: Maybe it would be worth logging the completion. Unless the task does it.

rkondratowicz: done

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -0,0 +1,105 @@
...
+
+        try {
+            BlobClient sourceBlob = sourceContainer.getBlobClient(envelope.fileName);
+            BlobClient targetBlob = targetContainer.getBlobClient(envelope.fileName);
+
+            if (!sourceBlob.exists()) {
+                logger.error("File already deleted. " + loggingContext);
+                envelopeRepository.markAsDeleted(envelope.id);
+            } else {
+                byte[] blobContent = download(sourceBlob);
```
doncem: hm. Since it's a different process and we are only moving blobs from one container to another, there is no need to download and then upload it somewhere. How about copying instead and then deleting after success?

PR #186, opened 2020-01-29
Add scheduled job to handle rejected files

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1003

chart changes: https://github.com/hmcts/blob-router-service/pull/191
flux changes: hmcts/cnp-flux-config#1966


charts/reform-scan-blob-router/values.yaml
```diff
@@ -28,6 +28,7 @@ java:
     FLYWAY_SKIP_MIGRATIONS: true
 
     DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
+    HANDLE_REJECTED_FILES_CRON: "0 0 7 * * *"
```
doncem: Changes to `values.yaml` requires chart version bump + separate PR + flux PR

nitinprabhu:
> chart changes better go with separate PR,
> bump the version,
> do we need to override new var in flux for AAT and/or prod?

good point. chart changes it would be nice if chart changes goes in separate PR plus flux changes. Then we can update tests/code.

rkondratowicz:
Flux PR: https://github.com/hmcts/cnp-flux-config/pull/1966

doncem:
> Flux PR: [hmcts/cnp-flux-config#1966](https://github.com/hmcts/cnp-flux-config/pull/1966)

Still config related changes should be as separate PR 😉

rkondratowicz:
> Still config related changes should be as separate PR 😉

ok, on it

rkondratowicz:
chart config changes:
https://github.com/hmcts/blob-router-service/pull/188

rkondratowicz:
All ready now

PR #187, opened 2020-01-29
Overwrite duplicate rejected files

rkondratowicz:
Next PR will introduce snapshots.


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -98,7 +98,8 @@ private void upload(BlobClient blobClient, byte[] blobContent, String loggingCon
             .getBlockBlobClient()
             .upload(
                 new ByteArrayInputStream(blobContent),
-                blobContent.length
+                blobContent.length,
+                true
```
rkondratowicz: `public BlockBlobItem upload(InputStream data, long length, boolean overwrite)`

rkondratowicz: throws an exception otherwise

PR #188, opened 2020-01-30
Add config changes for rejected files scheduled job

rkondratowicz:
for https://github.com/hmcts/blob-router-service/pull/186

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -28,6 +28,7 @@ java:
     FLYWAY_SKIP_MIGRATIONS: true
 
     DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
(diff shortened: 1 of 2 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -28,6 +28,7 @@ java:
     FLYWAY_SKIP_MIGRATIONS: true
 
     DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
+    HANDLE_REJECTED_FILES_CRON: "0/30 * * * * *"
```
doncem: Wasn't value different initially when I saw? So which one we are going for the main template :)

rkondratowicz: yeah, following existing convention now. Prod values are set in flux.

doncem: ah yea, of course :)

PR #194, opened 2020-01-30
Fix API GW tests

nitinprabhu:


src/smokeTest/resources/test-certificates.md
```diff
@@ -0,0 +1,34 @@
...
+
+Only valid certificates are stored in key vault.
+As inspected in test suite value is base64 encoded.
+Here is the command line which achieved desired outcome:
+
+```bash
+openssl base64 -in <infile> -out <outfile>
+```
+
+- `<infile>` would be certificate, e.g. `valid-certificate.jks`
```
doncem:
```suggestion
- `<infile>` would be certificate, e.g. `valid-certificate.pfx`
```

Jenkinsfile_CNP
```diff
@@ -14,27 +14,38 @@ def component = "blob-router"
...
+
+     secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
+
+     secret('storage-account-secondary-key', 'TEST_STORAGE_ACCOUNT_KEY'),
+     secret('storage-account-name', 'TEST_STORAGE_ACCOUNT_NAME'),
+
+     secret('crime-storage-connection-string', 'STORAGE_CRIME_CONNECTION_STRING'),
+     secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
+     secret('test-valid-key-store', 'TEST_VALID_KEY_STORE'),
+     secret('test-valid-key-store-password', 'TEST_VALID_KEY_STORE_PASSWORD')
```
doncem: those are already defined

lgonczar:
We need two more tests - for not yet valid certificate and for an expired one. That can be done later, as they don't exist in the processor, either.

Jenkinsfile_CNP
```diff
@@ -13,28 +13,36 @@ def component = "blob-router"
...
+
+     secret('crime-storage-connection-string', 'STORAGE_CRIME_CONNECTION_STRING'),
+     secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
+     secret('test-valid-key-store', 'TEST_VALID_KEY_STORE'),
+     secret('test-valid-key-store-password', 'TEST_VALID_KEY_STORE_PASSWORD'),
+
+     // smoke test secrets
+     secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
+     secret('test-valid-key-store', 'TEST_VALID_KEY_STORE'),
+     secret('test-valid-key-store-password', 'TEST_VALID_KEY_STORE_PASSWORD')
```
nitinprabhu: duplicates ?

lgonczar: It's either one set or another, from what I can see

Jenkinsfile_CNP
```diff
@@ -13,28 +13,28 @@ def component = "blob-router"
...
-    secret('crime-storage-connection-string', 'STORAGE_CRIME_CONNECTION_STRING')
-  ]
-]
+ def smokeTestSecrets = [
+   'reform-scan-${env}': [
+     secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
+     secret('test-valid-key-store', 'TEST_VALID_KEY_STORE'),
+     secret('test-valid-key-store-password', 'TEST_VALID_KEY_STORE_PASSWORD')
+   ]
+ ]
```
doncem: ah so that's what changed the block - extra space

PR #198, opened 2020-01-31
Add logging for the envelope id created for rejected blob

aliveni:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
@@ -105,7 +105,6 @@ public class BlobProcessor {
                         envelopeId
                     );
                 } else {
-                    logger.error("Invalid signature. Blob name: {} container: {}", blobName, containerName);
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -105,13 +105,17 @@ private void processBlob(String blobName, String containerName) {
...
+                    UUID envelopeId = envelopeRepository.insert(createNewEnvelope(
                         blobName,
                         containerName,
                         blobCreationDate,
                         Status.REJECTED
                     ));
+
+                    logger.info(
+                        "Rejected Blob. Name {} container {}. Envelope ID: {}", blobName, containerName, envelopeId
+                    );
```
doncem: perhaps we can combine into one log entry? not sure how precious are super edge case scenario people but failing to insert should rarely happen

aliveni: updated

PR #201, opened 2020-02-03
Add helper class for checking if rejected blob can be deleted

rkondratowicz:
Extracted from a bigger set of changes...

This class will be used by a job removing old rejected file from `*-rejected` containers

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/RejectedBlobChecker.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/RejectedBlobChecker.java
@@ -1,0 +1,27 @@
+package uk.gov.hmcts.reform.blobrouter.services;
+
+import com.azure.storage.blob.models.BlobItem;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/RejectedBlobChecker.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.hmcts.reform.blobrouter.services;
+
+import com.azure.storage.blob.models.BlobItem;
+import org.apache.commons.lang3.Validate;
+
+import java.time.Duration;
+
+import static java.time.OffsetDateTime.now;
+
+public class RejectedBlobChecker {
```
lgonczar: Should this be a service?

rkondratowicz: You mean the annotation?
Not yet. No config in app yaml etc.

PR #202, opened 2020-02-03
Fix logging in RejectedFilesHandler

rkondratowicz:
add cause


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -88,7 +88,7 @@ private void handleEnvelope(
                 logger.info("Rejected file successfully handled. " + loggingContext);
             }
         } catch (Exception exc) {
-            logger.error("Error handling rejected file. " + loggingContext);
+            logger.error("Error handling rejected file. " + loggingContext, exc);
```
doncem: just seeing this. in logging statements we should use it's own string interpolation. e.g. `log.level("some string {}", param, exception);`

rkondratowicz: fixed

doncem: and the other 2 instances in this method?

rkondratowicz: I'll change that separately.

PR #212, opened 2020-02-04
Remove apiGatewayTest module

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1024

### Change description ###

Remove apiGatewayTest module. Tests from that module have been moved to smoke tests.

```
[ ] Yes
[x] No
```


README.md
```diff
@@ -129,12 +129,6 @@ curl -v --key private.pem --cert cert.pem https://core-api-mgmt-{environment}.az
 
 You should get a response with status 200 and a token in the body.
 
-### API tests
-
-Jenkins (pipeline) runs the API gateway tests by executing `apiGateway` gradle task. This happens because
-there's a call to `enableApiGatewayTest()` in [Jenkins file](/Jenkinsfile_CNP). API tests
-are located in [apiGatewayTest](/src/apiGatewayTest/java/uk/gov/hmcts/reform/blobrouter) directory.
-
```
lgonczar: Removed this section, because API tests aren't anything special now and they're just a subset of smoke tests.

Jenkinsfile_CNP
```diff
@@ -83,7 +83,6 @@ withPipeline(type, product, component) {
   enableAksStagingDeployment()
   installCharts()
   enableSlackNotifications(channel)
-  enableApiGatewayTest()
```
nitinprabhu: worth commenting it rather removing as platform will fix the bug and we might forget ?

lgonczar:
It's not great to keep unused code and config (and there's also duplication). We could have a story to bring this change back when/if the API gateway test step is fixed, with reference to this PR.

Can close the PR if you feel strongly about it.

nitinprabhu:
> I don't think it's great if we keep unused code and config (and there's also duplication). We could have a story to bring this change back when/if the API gateway test step is fixed, with reference to this PR.
> 
> I can close the PR if you feel strongly about it.

agreed if we make sure that we add this back when platform bug is fixed then should not be a problem

lgonczar:
I've created the following story: https://tools.hmcts.net/jira/browse/BPS-1037

PR #217, opened 2020-02-05
add notification connection configurations

mdayican:
### Change description ###

add notification connection configurations -chart changes

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.preview.template.yaml
+++ b/charts/reform-scan-blob-router/values.preview.template.yaml
@@ -15,6 +15,9 @@ java:
     STORAGE_CRIME_ACCOUNT_KEY:
       secretRef: storage-secret-crime-{{ .Release.Name }}
       key: accessKey
+    SB_CONN_STRING:
+      secretRef: servicebus-secret-namespace-{{ .Release.Name }}
+      key: connectionString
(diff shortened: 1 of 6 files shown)


charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -81,3 +86,13 @@ crimeblobstorage:
   setup:
     containers:
       - bs-sit-scans-received
+
+servicebus:
+  resourceGroup: bulk-scan-aks
+  teamName: "Software Engineering"
```
rkondratowicz: ?

src/main/resources/bootstrap.yaml
```diff
@@ -16,3 +16,4 @@ spring:
         crime-storage-connection-string: storage.crime.connection-string
         flyway-password: flyway.password
         blob-router-POSTGRES-PASS: DB_PASSWORD
+        notifications-queue-send-connection-string: queue-notification-send
```
nitinprabhu: queue-notification-send -> queue-notification-send-conn-string ?

mdayican: changed

PR #218, opened 2020-02-05
Add integration test for rejected container cleaner

rkondratowicz:


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedContainerCleanerTest.java
```diff
@@ -0,0 +1,52 @@
...
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class RejectedContainerCleanerTest extends BlobStorageBaseTest {
+
+    @Mock RejectedBlobChecker blobChecker;
+
+    @Test
+    void should_delete_files_from_rejected_container() {
+        // given
+        var normalContainer = createContainer("sample-container");
+        var rejectedContainer = createContainer("sample-container-rejected");
```
nitinprabhu: would be nice to use actual object as it is not clear what object type is returned from createContainer ?

rkondratowicz: True... but it's a test ;)
OK, I'll remove `var`

rkondratowicz: done

PR #219, opened 2020-02-05
Move BlobReadinessChecker from BlobProcessor to ContainerProcessor

lgonczar:
### Change description ###

This pull request is a refactoring proposal.

Move BlobReadinessChecker from BlobProcessor to ContainerProcessor. The reason behind it is that BlobProcessor has too many dependencies and responsibilities, and it needs to lose some of them in order to be properly maintainable and easier to test.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
@@ -2,8 +2,12 @@ package uk.gov.hmcts.reform.blobrouter.tasks.processors;
 
 import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.BlobServiceClient;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -31,11 +38,25 @@ public void process(String containerName) {
 
             containerClient
                 .listBlobs()
-                .forEach(blob -> blobProcessor.process(blob.getName(), containerName));
+                .forEach(blob -> processIfReady(blob, containerName));
```
rkondratowicz: adding `.filter()` seems more natural

lgonczar: Filter was my first approach, but I went for this one because we're logging the fact of skipping the file, and having that in the predicate didn't seem right.

PR #175, opened 2020-01-28
Enable functional test for bulk scan blob transfer

doncem:
### JIRA link (if applicable) ###

[Create blob processing component](https://tools.hmcts.net/jira/browse/BPS-919)

### Change description ###

Introduced in #161

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #221, opened 2020-02-05
Enable functional tests for dispatching blobs to Bulk Scan

lgonczar:

### Change description ###

This PR is the same as https://github.com/hmcts/blob-router-service/pull/175, plus the last commit, fixing a JSON-related assertion.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProvider.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProvider.java
@@ -31,6 +31,7 @@ public class BlobContainerClientProvider {
             case BULKSCAN:
                 Configuration config = new Configuration();
                 config.put(Configuration.PROPERTY_HTTPS_PROXY, "proxyout.reform.hmcts.net:8080");
+                config.put(Configuration.PROPERTY_HTTP_PROXY, "proxyout.reform.hmcts.net:8080");
                 // retrieving a SAS token every time we're getting a client, but this will be cached in the future
                 return new BlobContainerClientBuilder()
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProvider.java
```diff
@@ -28,10 +29,14 @@ public BlobContainerClientProvider(
     public BlobContainerClient get(TargetStorageAccount targetStorageAccount, String containerName) {
         switch (targetStorageAccount) {
             case BULKSCAN:
+                Configuration config = new Configuration();
+                config.put(Configuration.PROPERTY_HTTPS_PROXY, "proxyout.reform.hmcts.net:8080");
+                config.put(Configuration.PROPERTY_HTTP_PROXY, "proxyout.reform.hmcts.net:8080");
```
rkondratowicz: is this needed both here and in application main()?

nitinprabhu: We should be able to remove all proxy config and test

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProvider.java
```diff
@@ -28,10 +29,14 @@ public BlobContainerClientProvider(
     public BlobContainerClient get(TargetStorageAccount targetStorageAccount, String containerName) {
         switch (targetStorageAccount) {
             case BULKSCAN:
+                Configuration config = new Configuration();
```
rkondratowicz: quite a new thing for us, so just reminding we _can_ use `var`

PR #223, opened 2020-02-06
Revert notification change

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1008

### Change description ###

processor needs event id to  handle notifications so inserting messages to processor queue would not solve the problem easily. We decide not to send notification until new notification micro service gets ready.

reverting conf changes: https://github.com/hmcts/blob-router-service/pull/217

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/requirements.yaml
+++ b/charts/reform-scan-blob-router/requirements.yaml
@@ -17,7 +17,3 @@ dependencies:
     version: ~0.2.1
     repository: '@hmctspublic'
     condition: crimeblobstorage.enabled
-  - name: servicebus
-    version: ~0.2.2
-    repository: '@hmctspublic'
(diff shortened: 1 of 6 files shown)


charts/reform-scan-blob-router/requirements.yaml
```diff
@@ -17,7 +17,3 @@ dependencies:
     version: ~0.2.1
     repository: '@hmctspublic'
     condition: crimeblobstorage.enabled
-  - name: servicebus
-    version: ~0.2.2
-    repository: '@hmctspublic'
-    condition: servicebus.enabled
```
doncem: 😌

mdayican:
> any change in `Jenkins_CNP` needed?

no.
reverting #217

PR #228, opened 2020-02-07
Added find by date to EnvelopeRepository

apereverzin:

### Change description ###
Added 'find by date' to EnvelopeRepository. This 'find' method will be called from the service which will convert the report date into the datetime interval.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
@@ -43,7 +43,7 @@ public class EnvelopeRepositoryTest {
             "hello.zip",
             now(),
             now().plusSeconds(100),
-            DISPATCHED
+            Status.DISPATCHED
         );
 
         // when
@@ -166,7 +166,7 @@ public class EnvelopeRepositoryTest {
         final String container = "bar";
 
         // and
-        repo.insert(new NewEnvelope(container, fileName, now(), now(), DISPATCHED));
+        repo.insert(new NewEnvelope(container, fileName, now(), now(), Status.DISPATCHED));
 
         // when
         Optional<Envelope> result = repo.find(fileName, container);
@@ -181,7 +181,7 @@ public class EnvelopeRepositoryTest {
     @Test
     void should_return_empty_optional_when_envelope_for_given_container_and_file_name_does_not_exist() {
         // given
-        repo.insert(new NewEnvelope("a", "b", now(), now(), DISPATCHED));
+        repo.insert(new NewEnvelope("a", "b", now(), now(), Status.DISPATCHED));
 
         // when
         Optional<Envelope> result = repo.find("some_other_file_name", "some_other_container");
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepository.java
```diff
@@ -75,6 +76,16 @@ public EnvelopeRepository(NamedParameterJdbcTemplate jdbcTemplate, EnvelopeMappe
         );
     }
 
+    public List<Envelope> find(Instant from, Instant to) {
```
rkondratowicz: maybe all we need is one parameter - date? (without the time)

lgonczar:
Maybe we should prepare the report based on events. Otherwise, we won't include the files that we failed to processed. Worth having a discussion.

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepository.java
apereverzin: One parameter - date - will be in a service method, this method will convert the date into the from - to range

rkondratowicz: if you think we'll need range somewhere else...

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -265,7 +266,7 @@ private void blobExists(OffsetDateTime time, byte[] contentToDownload) {
         }
 
         given(blobProperties.getLastModified()).willReturn(time);
-        given(envelopeRepo.find(any(), any())).willReturn(Optional.empty());
+        given(envelopeRepo.find(anyString(), anyString())).willReturn(Optional.empty());
```
rkondratowicz: unrelated

apereverzin: It is a fix for a compilation error :(

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
```diff
@@ -183,6 +189,32 @@ void should_return_empty_optional_when_envelope_for_given_container_and_file_nam
...
+        // given
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+        Instant instant1 = LocalDateTime.parse("2019-12-19 10:31:25", formatter).toInstant(ZoneOffset.UTC);
+        Instant instant2 = LocalDateTime.parse("2019-12-20 11:32:26", formatter).toInstant(ZoneOffset.UTC);
+        Instant instant3 = LocalDateTime.parse("2019-12-20 12:33:27", formatter).toInstant(ZoneOffset.UTC);
+        Instant instant4 = LocalDateTime.parse("2019-12-21 13:34:28", formatter).toInstant(ZoneOffset.UTC);
+        repo.insert(new NewEnvelope("a", "b", instant1, instant1, Status.DISPATCHED));
+        repo.insert(new NewEnvelope("c", "d", instant2, instant2, Status.DISPATCHED));
+        repo.insert(new NewEnvelope("e", "f", instant3, instant3, Status.DISPATCHED));
+        repo.insert(new NewEnvelope("g", "h", instant4, instant4, Status.DISPATCHED));
```
rkondratowicz: you've set the same instant in two fields of envelope, test name or test data don't explain which property of envelope is used for search

rkondratowicz:
> Maybe we should prepare the report based on events. Otherwise, we won't include the files that we failed to process. Worth having a discussion.

@lgonczar Maybe we should introduce a new status and create an envelope?

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
rkondratowicz: oh, you've added a new method in repo with signature similar to already existing one

rkondratowicz:
Now that events repository exists we can use it instead

rkondratowicz:
@apereverzin can we close this PR?

PR #231, opened 2020-02-07
use BlockBlobClient

mdayican:
### Change description ###

use BlockBlobClient instead of non-blocking for downloading

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
@@ -64,9 +64,10 @@ public class RejectedFilesHandler {
         Envelope envelope
     ) {
         String loggingContext = String.format(
-            "File name: %s. Container: %s",
+            "File name: %s. Source Container: %s. Target Container: %s",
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -94,7 +95,7 @@ private void handleEnvelope(
 
     private byte[] download(BlobClient blobClient) throws IOException {
         try (var outputStream = new ByteArrayOutputStream()) {
-            blobClient.download(outputStream);
+            blobClient.getBlockBlobClient().download(outputStream);
```
doncem: I still think we should refactor this approach into `copy` functionality straight from one container to another

mdayican: definitely this can cause out of memory

doncem: should be considered as a separate PR ;). let's fix this flow

PR #232, opened 2020-02-07
Clean up zip verification

rkondratowicz:
- throw more appropriate exception when the content of zip file is invalid
- use already existing constant for zip file name in processor

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
@@ -17,6 +17,7 @@ import uk.gov.hmcts.reform.blobrouter.exceptions.InvalidZipArchiveException;
 import uk.gov.hmcts.reform.blobrouter.services.BlobSignatureVerifier;
 import uk.gov.hmcts.reform.blobrouter.services.storage.BlobDispatcher;
 import uk.gov.hmcts.reform.blobrouter.services.storage.LeaseClientProvider;
+import uk.gov.hmcts.reform.blobrouter.util.ZipVerifiers;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -163,13 +161,13 @@ private void processBlob(String blobName, String containerName) {
             ZipEntry entry = null;
 
             while ((entry = zipStream.getNextEntry()) != null) {
-                if (ENVELOPE_ENTRY_NAME.equals(entry.getName())) {
+                if (ZipVerifiers.DOCUMENTS_ZIP.equals(entry.getName())) {
```
nitinprabhu: `ZipVerifiers.ENVELOPES_ZIP` may be ?

rkondratowicz: For now I'm just removing the duplication of contants.

The other name is not great either, currently there are those constants in `ZipVerifiers` class:

```java
    public static final String DOCUMENTS_ZIP = "envelope.zip";
    public static final String SIGNATURE_SIG = "signature";
    public static final String INVALID_SIGNATURE_MESSAGE = "Zip signature failed verification";
```

I'd probably move the file names constants to a separate class (like `ZipFileEntries`) and call them `ENVELOPE_ZIP` and `SIGNATURE` respectively.

lgonczar:
Will `InvalidZipArchiveException` cause rejection?

rkondratowicz:
> Will `InvalidZipArchiveException` cause rejection?

good point. Added extra `catch` and new tests

PR #233, opened 2020-02-07
Email sender

apereverzin:

### Change description ###
EmailSender class sending email with attachments. Code is partially copied from bulkscan-processor.

EmailSender will be used for sending reports.

In this PR Spring annotations are excluded from EmailSender to prevent any emails caused by a corrupted configuration. In the next PR there will be Spring annotations and integration tests.

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -228,6 +228,10 @@ dependencies {
 
   implementation group: 'uk.gov.hmcts.reform', name: 'properties-volume-spring-boot-starter', version:'0.0.4'
 
+  implementation group: 'org.springframework', name: 'spring-context-support', version: '5.2.2.RELEASE'
+
+  implementation group: 'com.sun.mail', name: 'jakarta.mail', version: '1.6.3'
+
```
doncem: Unusual to our setup comparing with previous integrations of reports. Can you elaborate more around this choice?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/email/EmailSender.java
```diff
@@ -0,0 +1,52 @@
...
+    private final JavaMailSender mailSender;
+
+    // region constructor
+    public EmailSender(JavaMailSender mailSender) {
+        this.mailSender = mailSender;
+    }
+    // endregion
+
+    public void sendMessageWithAttachments(
+        String subj,
```
doncem: isn't it `subject`?

build.gradle
```diff
@@ -228,6 +229,8 @@ dependencies {
 
   implementation group: 'uk.gov.hmcts.reform', name: 'properties-volume-spring-boot-starter', version:'0.0.4'
 
+  implementation group: 'org.springframework', name: 'spring-context-support', version: '5.2.3.RELEASE'
```
doncem: now this is probably redundant

apereverzin: yes, indeed. have a nice weekend

src/main/java/uk/gov/hmcts/reform/blobrouter/services/email/EmailSender.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.mail.javamail.JavaMailSender;
+import org.springframework.mail.javamail.MimeMessageHelper;
+import org.springframework.stereotype.Component;
+
+import java.io.File;
+import java.util.Map;
+import javax.mail.internet.MimeMessage;
+
+@Component
+@ConditionalOnProperty(prefix = "spring.mail", name = "host")
```
doncem: probably i'm missing the big picture or those zips in diff are truly blocking me.. i don't see any config change. perhaps we don't need any components for now nor conditions. otherwise i would enquire integration test to assure email is never sent on any environment other than prod

apereverzin: zip files are for testing only. They are really not zip.
I will deal with integration testing, I agree unit tests are not enough.

rkondratowicz:
worth putting those files from resources used in tests into their own folder imho

PR #239, opened 2020-02-10
use custom connection string

mdayican:
### Change description ###

use custom connection string - in the connection string we will use the app gateway url
"storage-reform-connection-string" value added to keyvault. 

flux change: https://github.com/hmcts/cnp-flux-config/pull/2078

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
```diff
@@ -32,11 +32,10 @@ public LeaseClientProvider getLeaseClientProvider() {
     @Bean
     public BlobServiceClient getStorageClient(
         StorageSharedKeyCredential credentials,
-        @Value("${storage.url}") String storageUrl
+        @Value("${storage.reform.connection-string}") String connectionString
```
doncem: how will that work for all the other environments?

mdayican: for aat and  prod we will use external url which will be bootstrap from vault

doncem: in that case, we need all the usual recipes for value: chart versions, values template for all envs, flux config 😉

charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -41,6 +41,7 @@ java:
     STORAGE_BULKSCAN_URL: https://bulkscanaat.blob.core.windows.net
 
     STORAGE_URL: "https://$(STORAGE_ACCOUNT_NAME).blob.core.windows.net"
+    STORAGE_REFORM_CONNECTION_STRING: "DefaultEndpointsProtocol=https;BlobEndpoint=https://$(STORAGE_ACCOUNT_NAME).blob.core.windows.net;AccountName=$(STORAGE_ACCOUNT_NAME);AccountKey=$(STORAGE_ACCOUNT_KEY);"
```
nitinprabhu: do you intend to use internal or app gw url ?

mdayican: for  pr  it is internal

doncem: aye, `values.yaml` should use app gw whereas preview is fine here

src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
mdayican: I will add values to vault and revert code changes and release this.

charts/reform-scan-blob-router/values.preview.template.yaml
nitinprabhu: `STORAGE_REFORM_CONNECTION_STRING` nothing in values yaml ?

mdayican: - storage-reform-connection-string

charts/reform-scan-blob-router/values.yaml
```diff
@@ -16,6 +16,7 @@ java:
         - crime-storage-connection-string
         - flyway-password
         - blob-router-POSTGRES-PASS
+        - storage-reform-connection-string
```
doncem: still missing value in prod vault

mdayican: @nitinprabhu  will add

mdayican: I added

PR #204, opened 2020-02-03
Track schedules as requests in App Insights

doncem:
### Change description ###

Code was pending locally until #160 comes back to play. Now possible to add next level - top level tracking for e2e transaction feature provided by App Insights.

References:
- [microsoft guideline](https://github.com/microsoft/ApplicationInsights-Java/wiki/Distributed-Tracing-in-Asynchronous-Java-Applications#context-propagation-in-scheduled-events)
- [send letter service](https://github.com/hmcts/send-letter-service/blob/master/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java) introduced [some time ago](https://github.com/hmcts/send-letter-service/pull/437)
- web filter fix for app insights dependency in integration tests hmcts/send-letter-service#563

Note. 2 extra commits "fixing" up modified test file

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #242, opened 2020-02-11
Generalize logging in scheduled jobs

rkondratowicz:
Now that app insights alerts for jobs rely on certain entries in logs, it is important to keep the logs consistent and not change them (unless alert is updated as well).

Hopefully after this change this will be less likely to happen in the future.

Note: not using inheritance intentionally.


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/utils/LoggingHelper.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.slf4j.Logger;
+
+public final class LoggingHelper {
+
+    private LoggingHelper() {
+        // util class
+    }
+
+    public static void wrapWithJobLog(Logger logger, String taskName, Runnable runnable) {
```
rkondratowicz: I'm open to suggestions on alternative name :)

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/utils/LoggingHelper.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmcts.reform.blobrouter.tasks.utils;
+
+import org.slf4j.Logger;
+
+public final class LoggingHelper {
```
nitinprabhu: this is also executing the tasks ? so probably `ScheduledTaskRunner` ?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/utils/LoggingHelper.java
nitinprabhu: `executeWithJobLog` ?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/utils/LoggingHelper.java
rkondratowicz: it's not running scheduled tasks.
Scheduled tasks are run automagically by Spring, then scheduled tasks use this class to surround code with logs.

lgonczar: Or maybe `JobRunner`?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/utils/LoggingHelper.java
lgonczar: I'd be even fine with `run`.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/utils/LoggingHelper.java
```diff
@@ -0,0 +1,16 @@
...
+
+    private LoggingHelper() {
+        // util class
+    }
+
+    public static void wrapWithJobLog(Logger logger, String taskName, Runnable runnable) {
+        logger.info("Started {} job", taskName); // DO NOT change this message. App insights alerts are based on it.
+        runnable.run();
+        logger.info("Finished {} job", taskName);
```
lgonczar: Not sure, but maybe it's worth handling `Exception` here.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/utils/LoggingHelper.java
doncem: perhaps worth considering continuously developing/improving/bringing-up-to-speed app insights setup for the service by making #204 pass. Currently it's reporting that rejected tasks are not working properly. Same thing is also done in send letter service too. Then logging will be more trivial than ever. and will not involve in tampering schedules themselves

btw, @nitinprabhu, what you are referring perhaps closely looks like the #102, which is, again, setup for #204 thing to happen

nitinprabhu: `JobRunner` sounds good @lgonczar  @rkondratowicz

PR #243, opened 2020-02-11
Remove ZipStreamWithSignature class

rkondratowicz:
Instead of loading public key from file when creating instance of `ZipStreamWithSignature` (which bundles zip content and **public key**, not _signature_ as the name could suggest), simply load public key once in the service.

Hopefully clearer now.


src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobSignatureVerifier.java
```diff
@@ -4,33 +4,40 @@
...
 
-    private final String publicKeyDerFilename;
+    private final String publicKeyBase64;
 
     public BlobSignatureVerifier(
         @Value("${public-key-der-file}") String publicKeyDerFilename
     ) {
-        this.publicKeyDerFilename = publicKeyDerFilename;
+        try {
+            publicKeyBase64 = Base64.getEncoder().encodeToString(toByteArray(getResource(publicKeyDerFilename)));
```
doncem: is it ok time to start using `Base64.getMimeEncoder()` instead?

rkondratowicz: Let's leave 'RFC4648 vs RFC2045' for a separate PR ;)

src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobSignatureVerifier.java
```diff
@@ -4,33 +4,40 @@
...
+import static com.google.common.io.Resources.toByteArray;
 import static org.slf4j.LoggerFactory.getLogger;
 
 @Component
 public class BlobSignatureVerifier {
 
     private static final Logger logger = getLogger(BlobSignatureVerifier.class);
 
-    private final String publicKeyDerFilename;
+    private final String publicKeyBase64;
```
lgonczar: Maybe it would be worth adding a comment or sth to contain information about the key being a der file.

rkondratowicz: see constructor below

PR #244, opened 2020-02-11
Add events table

doncem:

[Record events in DB](https://tools.hmcts.net/jira/browse/BPS-1040)

### Change description ###

This PR includes migration script and models/repository accompanied with integration test. Following one should start populating the table.

The setup used for events table here is very closely mimicking envelopes entities, currently present in repository

Note. `EventsRecord` has extra constructor which does not include `notes` field to simplify its construction. Not all events need a note only failing ones we are interested in. Integrations test case takes into account that

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V004__Add_events_table.sql
```diff
@@ -0,0 +1,11 @@
...
+  id          BIGSERIAL     PRIMARY KEY,
+  container   VARCHAR(50)   NOT NULL,
+  file_name   VARCHAR(255)  NOT NULL,
+  created_at  TIMESTAMP     NOT NULL,
+  event       VARCHAR(100)  NOT NULL,
+  notes       TEXT          NULL
+);
+
+CREATE INDEX events_container_filename_idx ON events (container, file_name);
+CREATE INDEX events_container_event_idx ON events (container, event);
```
rkondratowicz: Seems a bit premature.
Why the second index?

doncem: Can remove as no need at the given state

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EventRecordRepositoryTest.java
```diff
@@ -0,0 +1,101 @@
...
+            .hasSize(1)
+            .first()
+            .satisfies(record -> {
+                assertThat(record.id).isEqualTo(id);
+                assertThat(record.event).isEqualTo(event);
+
+                if (notes == null) {
+                    assertThat(record.notes).isNull();
+                } else {
+                    assertThat(record.notes).isEqualTo(notes);
```
rkondratowicz: will this not work for null?

src/main/resources/db/migration/V004__Add_events_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE events (
+  id          BIGSERIAL     PRIMARY KEY,
```
rkondratowicz: 💪

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EventRecordRepositoryTest.java
doncem: nope, there's a null check in the method

src/main/resources/db/migration/V004__Add_events_table.sql
lgonczar: 👍 , we need to know the order of events

rkondratowicz: I was refering to BIG serial ;)

src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/Event.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.reform.blobrouter.data.model;
+
+public enum Event {
+    FILE_RECEIVED,
```
lgonczar: Not sure we can catch `FILE_RECEIVED` event. We'll either have DISPATCHED, REJECTED, RECEIVED_DUPLICATE or FAILED_TO_PROCESS as the first event.

src/main/resources/db/migration/V004__Add_events_table.sql
lgonczar: I know, my comment wasn't related

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EventRecordRepositoryTest.java
```diff
@@ -0,0 +1,96 @@
...
+            .hasSize(3)
+            .extracting(record -> tuple(record.event, record.notes))
+            .containsExactlyInAnyOrder(
+                tuple(FILE_RECEIVED, null),
+                tuple(DISPATCHED, null),
+                tuple(REJECTED, "description")
+            );
+    }
+
+    private static Stream<Arguments> eventWithNotes() {
```
lgonczar: How about `events`?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EventRecordRepositoryTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
+import static uk.gov.hmcts.reform.blobrouter.data.model.Event.DISPATCHED;
+import static uk.gov.hmcts.reform.blobrouter.data.model.Event.FILE_RECEIVED;
+import static uk.gov.hmcts.reform.blobrouter.data.model.Event.REJECTED;
+
+@ActiveProfiles({"integration-test", "db-test"})
+@SpringBootTest
+public class EventRecordRepositoryTest {
```
lgonczar: Worth testing if the envelope fetching method doesn't return envelopes from a different container or for a different file.

src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/Event.java
doncem: perhaps rename? what i meant was smth like `FILE_START_PROCESSING`. and can be inserted once processing checker has allowed to continue - don't want to end up with godzillion entries

lgonczar: Renaming sounds good

PR #248, opened 2020-02-12
set staging rejected file handle 10 sec

mdayican:
### Change description ###

Change staging rejected file handle every 10 sec
to see if it is changing the task running interval,

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.aat.template.yaml
```diff
@@ -3,6 +3,7 @@ java:
     CRIME_ENABLED: "true"
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
     DELETE_DISPATCHED_FILES_CRON: "0/10 * * * * *"
+    HANDLE_REJECTED_FILES_CRON: "0/10 * * * * *"
```
nitinprabhu: minimum overriding in this file is what we need to do as this file is only applicable to staging

PR #249, opened 2020-02-12
Don't check if rejected file was removed in functional tests

rkondratowicz:
Tests are passing on preview, but not on AAT (job not working, getting errors).
Unblocking pipeline.


rkondratowicz:
Problem fixed (infrastructure level), no need for this change

PR #253, opened 2020-02-13
instead of download and upload try to copy directly between containers

mdayican:
### Change description ###

instead of download and upload try to copy directly between containers
this may help memory usage and firewall content-type  issue

sas token required to copy
https://docs.microsoft.com/en-us/rest/api/storageservices/copy-file

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/BlobRejectTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/BlobRejectTest.java
@@ -1,5 +1,8 @@
 package uk.gov.hmcts.reform.blobrouter;
 
+import com.azure.storage.blob.BlobServiceClient;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.codec.digest.DigestUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -7,6 +10,7 @@ import java.util.concurrent.TimeUnit;
 
 import static com.jayway.awaitility.Awaitility.await;
 import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
 import static uk.gov.hmcts.reform.blobrouter.data.model.Status.REJECTED;
 import static uk.gov.hmcts.reform.blobrouter.envelope.ZipFileHelper.createZipArchive;
 import static uk.gov.hmcts.reform.blobrouter.storage.StorageHelper.blobExists;
@@ -42,6 +46,13 @@ public class BlobRejectTest extends FunctionalTestBase {
 
         // and
         assertFileInfoIsStored(fileName, config.crimeSourceContainer, REJECTED, true);
+
+        assertBlobIsPresentInStorage(
+            blobRouterStorageClient,
+            "crime-rejected",
+            fileName,
+            wrappingZipContent
+        );
     }
 
     @Test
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -93,27 +95,29 @@ private void handleEnvelope(
...
 
-        blobClient
-            .getBlockBlobClient()
-            .upload(
-                new ByteArrayInputStream(blobContent),
-                blobContent.length,
-                true
-            );
+        targetBlob
+            .getBlockBlobClient().copyFromUrl(sourceBlob.getBlobUrl() + "?" + createSas(sourceBlob,5));
```
doncem: maybe we can just set expiry locally in that method instead. if not then missing space here 😛

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/BlobRejectTest.java
```diff
@@ -42,6 +46,13 @@ void should_reject_crime_envelope_with_invalid_signature() throws Exception {
 
         // and
         assertFileInfoIsStored(fileName, config.crimeSourceContainer, REJECTED, true);
+
+        assertBlobIsPresentInStorage(
```
rkondratowicz: we may need to wait here for a bit as well

mdayican: I think first one should enough,  if it is not in the source container it should be in target

rkondratowicz: true!

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -93,27 +95,29 @@ private void handleEnvelope(
...
+        OffsetDateTime expiryTime =
+            OffsetDateTime.of(LocalDateTime.now().plus(5, ChronoUnit.MINUTES), ZoneOffset.UTC);
+
+        BlobServiceSasSignatureValues blobServiceSasSignatureValues =
+            new BlobServiceSasSignatureValues(
+                expiryTime,
+                new BlobContainerSasPermission().setReadPermission(true)
+            );
+
+        return sourceBlob.generateSas(blobServiceSasSignatureValues);
```
nitinprabhu: Sas should be for destination blob ? As you are copying to destination container ?

mdayican: sas is getting from source

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -93,27 +95,29 @@ private void handleEnvelope(
...
-            );
+        targetBlob
+            .getBlockBlobClient().copyFromUrl(sourceBlob.getBlobUrl() + "?" + createSas(sourceBlob));
+
         logger.info("File successfully uploaded to rejected container. " + loggingContext);
     }
+
+    private String createSas(BlobClient sourceBlob) {
+        OffsetDateTime expiryTime =
+            OffsetDateTime.of(LocalDateTime.now().plus(5, ChronoUnit.MINUTES), ZoneOffset.UTC);
```
mdayican: may there be time differences between aks and blob ?? , 5 minutes is  enough?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
```diff
@@ -93,27 +95,29 @@ private void handleEnvelope(
...
 
-        blobClient
-            .getBlockBlobClient()
-            .upload(
-                new ByteArrayInputStream(blobContent),
-                blobContent.length,
-                true
-            );
+        targetBlob
+            .getBlockBlobClient().copyFromUrl(sourceBlob.getBlobUrl() + "?" + createSas(sourceBlob));
```
nitinprabhu: when you log you use `sourceBlob.getBlockBlobClient().getBlobUrl()` ?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
nitinprabhu: both should be updated so should be fine

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandler.java
mdayican: chnaged

PR #258, opened 2020-02-14
Move method for finding enabled source containers

rkondratowicz:
From `BlobDispatcherTask ` to `ServiceConfiguration `.

The component  dealing with duplicates will need this as well.


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
```diff
@@ -39,20 +37,10 @@ public BlobDispatcherTask(
     public void run() {
         logger.info("Started {} job", TASK_NAME);
 
-        List<String> containers = getAvailableContainers();
+        List<String> containers = serviceConfiguration.getEnabledSourceContainers();
```
mdayican: do we need to get the list every time?

rkondratowicz: good point, I'll move it to constructor in next PR.

PR #260, opened 2020-02-14
Add component for finding duplicate envelopes

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1035

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/DuplicateFinder.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/DuplicateFinder.java
@@ -27,7 +27,7 @@ public class DuplicateFinder {
             .getBlobContainerClient(containerName)
             .listBlobs()
             .stream()
-            .map(blob -> envelopeService.findEnvelopes(blob.getName(), containerName))
+            .map(blob -> envelopeService.findEnvelope(blob.getName(), containerName))
             .flatMap(Optional::stream)
             .filter(envelope -> envelope.isDeleted)
             .collect(toList());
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/DuplicateFinder.java
```diff
@@ -0,0 +1,35 @@
...
+    public List<Envelope> findIn(String containerName) {
+        return storageClient
+            .getBlobContainerClient(containerName)
+            .listBlobs()
+            .stream()
+            .map(blob -> envelopeService.findEnvelopes(blob.getName(), containerName))
+            .flatMap(Optional::stream)
+            .filter(envelope -> envelope.isDeleted)
+            .collect(toList());
+    }
```
doncem: so this eventually replace the https://github.com/hmcts/blob-router-service/blob/master/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java#L63-L76?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/DuplicateFinder.java
```diff
@@ -0,0 +1,35 @@
...
+        this.storageClient = storageClient;
+        this.envelopeService = envelopeService;
+    }
+
+    public List<Envelope> findIn(String containerName) {
+        return storageClient
+            .getBlobContainerClient(containerName)
+            .listBlobs()
+            .stream()
+            .map(blob -> envelopeService.findEnvelopes(blob.getName(), containerName))
```
mdayican: may be directly get deleted files?

rkondratowicz: Method name is misleading, this is returning a single envelope.
Fix in https://github.com/hmcts/blob-router-service/pull/264

PR #266, opened 2020-02-17
Removed irrelevant profiles

apereverzin:

### Change description ###
Removed 'db-test- profile annotations from integration tests where these profiles are irrelevant.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
@@ -3,13 +3,16 @@ package uk.gov.hmcts.reform.blobrouter.tasks.processors;
 import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.models.BlobItem;
 import com.azure.storage.blob.specialized.BlobLeaseClientBuilder;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -26,7 +25,6 @@
 import static uk.gov.hmcts.reform.blobrouter.data.model.Status.DISPATCHED;
 import static uk.gov.hmcts.reform.blobrouter.testutils.DirectoryZipper.zipAndSignDir;
 
-@ActiveProfiles("db-test")
```
rkondratowicz: this test does use database

apereverzin: It does not need DbHelper - the only component linked to this profile.

rkondratowicz: perhaps this profile can be removed then (not only profile usage)

apereverzin: This profile makes sense, it is limiting instantiation of DbHelper.

lgonczar:
This change is fine, as it just removed unnecessary profile references.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
lgonczar: It would be best if this test did DB cleanup at the start. It's not strictly related with this PR, but it needs to be fixed. I think we need that profile reference here in the end, so that we can use DB Helper.

PR #269, opened 2020-02-17
Extract component for moving files to rejected container

rkondratowicz:
The new job that handles duplicate files will need to move duplicates to rejected container.

Extracting code doing that from already existing component.

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandlerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandlerTest.java
@@ -37,7 +37,7 @@ class RejectedFilesHandlerTest extends BlobStorageBaseTest {
     @BeforeEach
     void setUp() {
         dbHelper.deleteAll();
-        mover = new RejectedFilesHandler(storageClient, envelopeService);
+        mover = new RejectedFilesHandler(envelopeService, new BlobRejecter(storageClient));
     }
 
     @AfterEach
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandlerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandlerTest.java
@@ -37,7 +37,7 @@ class RejectedFilesHandlerTest extends BlobStorageBaseTest {
     @BeforeEach
     void setUp() {
         dbHelper.deleteAll();
-        mover = new RejectedFilesHandler(envelopeService, new BlobRejecter(storageClient));
+        mover = new RejectedFilesHandler(envelopeService, new BlobMover(storageClient));
     }
 
     @AfterEach
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
@@ -5,6 +5,7 @@ import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.sas.BlobContainerSasPermission;
 import com.azure.storage.blob.sas.BlobServiceSasSignatureValues;
 import org.slf4j.Logger;
+import org.springframework.stereotype.Component;
 
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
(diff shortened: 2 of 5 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/storage/RejectedFilesHandlerTest.java
```diff
@@ -37,7 +37,7 @@
     @BeforeEach
     void setUp() {
         dbHelper.deleteAll();
-        mover = new RejectedFilesHandler(storageClient, envelopeService);
+        mover = new RejectedFilesHandler(envelopeService, new BlobRejecter(storageClient));
```
rkondratowicz: This class no longer accesses blob storage directly.
I'll remove integration test for it and introduce a new one (for `BlobRejecter`) in the next PR.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public void moveToRejectedContainer(String blobName, String containerName) {
+
+        var sourceBlob = storageClient.getBlobContainerClient(containerName).getBlobClient(blobName);
+        var targetBlob = storageClient.getBlobContainerClient(containerName + REJECTED_CONTAINER_SUFFIX).getBlobClient(
+            blobName);
+
+        if (targetBlob.exists()) {
+            targetBlob.createSnapshot();
+        }
```
doncem: perhaps we should create snapshot right before copying the blob?

rkondratowicz: Good point!
Just moving existing code to a new class in this PR.
I'll fix this in a separate PR (and add a test - no need to create a snapshot if file in source container doesn't exist).

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
```diff
@@ -0,0 +1,72 @@
...
+
+        String loggingContext = String.format(
+            "File name: %s. Source Container: %s. Target Container: %s",
+            blobName,
+            sourceBlob.getContainerName(),
+            targetBlob.getContainerName()
+        );
+
+        if (!sourceBlob.exists()) {
+            logger.error("File already deleted. " + loggingContext);
```
lgonczar: Maybe worth saying a bit more, i.e. that the file was supposed to be moved to rejected container.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
```diff
@@ -0,0 +1,72 @@
...
+        } else {
+            copy(targetBlob, sourceBlob, loggingContext);
+            sourceBlob.delete();
+            logger.info("File successfully moved to rejected container. " + loggingContext);
```
lgonczar: After `copy` we know that the file has been copied. Maybe in this last `info` we should say that the source file was deleted.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
```diff
@@ -0,0 +1,72 @@
...
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
+
+import static org.slf4j.LoggerFactory.getLogger;
+import static uk.gov.hmcts.reform.blobrouter.services.storage.RejectedFilesHandler.REJECTED_CONTAINER_SUFFIX;
+
+@Component
+public class BlobRejecter {
```
lgonczar: The blob is rejected earlier and this tool just moves it. Maybe `RejectedBlobMover` or something like that?

rkondratowicz: Note that the duplicate handler will use this class as well

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
rkondratowicz: moved existing code

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
rkondratowicz: moved existing code

lgonczar:
Nice idea. It would be good to have separate tests for this new class, even if they are integration tests.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobRejecter.java
lgonczar: So `BlobMover`?

rkondratowicz: renamed to `BlobMover`, method is called `moveToRejectedContainer`

PR #265, opened 2020-02-17
Record events to the database

doncem:
### JIRA link (if applicable) ###

[Record events in DB](https://tools.hmcts.net/jira/browse/BPS-1040)

### Change description ###

Comes in with slight boilerplate with extra testing. What got me was the deletion status: one from source another from rejected container. Handled those appropriately

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #271, opened 2020-02-17
Add service method for saving 'duplicate envelope' event

rkondratowicz:
Following naming convention from https://github.com/hmcts/blob-router-service/pull/265


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -68,4 +70,9 @@ private UUID createEnvelope(String containerName, String blobName, Instant blobC
     public void markEnvelopeAsDeleted(UUID envelopeId) {
         envelopeRepository.markAsDeleted(envelopeId);
     }
+
+    @Transactional
```
lgonczar: I wonder if this needs to be transactional.

PR #277, opened 2020-02-19
Store rejection reason in envelope events

rkondratowicz:
File can be rejected for several reasons.
Updated the `BlobVerifier` to return reason info in the verification result.

Now this reason is used when setting notes in the `REJECTED` envelope event.


src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobVerifier.java
```diff
@@ -35,20 +37,38 @@ public BlobVerifier(
...
-            return false;
+            return error("Invalid signature");
         } catch (InvalidZipArchiveException ex) {
             logger.info("Invalid zip archive. Blob name: {}", blobName, ex);
-            return false;
+            return error("Invalid zip archive");
         } catch (IOException ex) {
             logger.info("Error occurred when verifying file. Blob name: {}", blobName, ex);
-            return false;
+            return error(null);
```
lgonczar: Do we distinguish between errors with message and errors with null? This last one is some generic failure.

rkondratowicz: How about `"Error loading file"`?

lgonczar: I wonder if under this IOException there might be a recoverable failure, not strictly related with the file. Shouldn't we treat it as Exception?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobVerifier.java
```diff
@@ -35,20 +37,38 @@ public BlobVerifier(
...
+        }
+    }
+
+    public static class VerificationResult {
+        public final boolean isOk;
+        public final String error;
+
+        public VerificationResult(boolean isOk, String error) {
```
apereverzin: Should be private?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobVerifier.java
rkondratowicz: done in new PR: https://github.com/hmcts/blob-router-service/pull/279

src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobVerifier.java
rkondratowicz: done

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -124,15 +123,15 @@ private void processBlob(String blobName, String containerName) {
...
                     blobName,
                     blobCreationDate,
-                    INVALID_SIGNATURE
+                    verificationResult.error
                 );
 
                 logger.error(
-                    "{}. Rejected Blob. File name: {}, Container {}, New envelope ID: {}",
-                    INVALID_SIGNATURE,
+                    "Rejected Blob. File name: {}, Container: {}, New envelope ID: {}, Reason: {}",
```
lgonczar: 👍

PR #278, opened 2020-02-19
Add events to envelope info response

doncem:
### Change description ###

Since events are now recorded - information can be retrieved via endpoint. Adding `JsonProperty` to all places for consistency

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -2,10 +2,13 @@
 
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
+import reactor.util.function.Tuple2;
+import reactor.util.function.Tuples;
```
apereverzin: Good!

PR #279, opened 2020-02-19
Retry processing file if it cannot be loaded

rkondratowicz:
instead of rejecting it.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobVerifier.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobVerifier.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.reform.blobrouter.services;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.blobrouter.exceptions.BlobProcessingException;
 import uk.gov.hmcts.reform.blobrouter.exceptions.DocSignatureFailureException;
 import uk.gov.hmcts.reform.blobrouter.exceptions.InvalidConfigException;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -64,8 +66,8 @@
 
     @Mock BlobServiceClient blobServiceClient;
     @Mock BlobContainerClient containerClient;
-    @Mock BlobClient blobClient;
-    @Mock BlobProperties blobProperties;
+    @Mock(lenient = true) BlobClient blobClient;
+    @Mock(lenient = true) BlobProperties blobProperties;
```
rkondratowicz: not all stubbing from helper methods is now necessary

src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobVerifier.java
```diff
@@ -47,8 +48,10 @@ public boolean verifyZip(String blobName, byte[] rawBlob) {
             logger.info("Invalid zip archive. Blob name: {}", blobName, ex);
             return false;
         } catch (IOException ex) {
-            logger.info("Error occurred when verifying file. Blob name: {}", blobName, ex);
-            return false;
+            throw new BlobProcessingException(
```
lgonczar: Should this happen for all other exceptions, i.e. should we change the exception type in this `catch` to `Exception`?

lgonczar: Plus, if we make it `Exception`, we can test that `BlobVerifier` throws `BlobProcessingExceptions` when it fails to verify the file.

PR #281, opened 2020-02-20
Repo to query envelopes per day

apereverzin:

### Change description ###
EnvelopeSummary and EnvelopeSummaryRepository.

EnvelopeSummaryRepository will be called from the service layer which will be in charge of deriving 'from' and 'to' timestamps for a given date.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
@@ -6,7 +6,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.ActiveProfiles;
 import uk.gov.hmcts.reform.blobrouter.data.model.EnvelopeSummary;
-import uk.gov.hmcts.reform.blobrouter.data.model.EventRecord;
 import uk.gov.hmcts.reform.blobrouter.data.model.NewEnvelope;
 import uk.gov.hmcts.reform.blobrouter.data.model.NewEventRecord;
 import uk.gov.hmcts.reform.blobrouter.data.model.Status;
@@ -16,25 +15,21 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.List;
-import java.util.UUID;
 
 import static java.time.ZoneOffset.UTC;
 import static java.util.Comparator.comparingLong;
 import static java.util.stream.Collectors.toList;
 import static org.assertj.core.api.Assertions.assertThat;
-import static uk.gov.hmcts.reform.blobrouter.data.model.Event.DELETED;
-import static uk.gov.hmcts.reform.blobrouter.data.model.Event.DELETED_FROM_REJECTED;
 import static uk.gov.hmcts.reform.blobrouter.data.model.Event.DISPATCHED;
 import static uk.gov.hmcts.reform.blobrouter.data.model.Event.DUPLICATE_REJECTED;
 import static uk.gov.hmcts.reform.blobrouter.data.model.Event.FILE_PROCESSING_STARTED;
-import static uk.gov.hmcts.reform.blobrouter.data.model.Event.REJECTED;
 
 @ActiveProfiles({"integration-test", "db-test"})
 @SpringBootTest
 public class EnvelopeSummaryRepositoryTest {
     @Autowired private EnvelopeRepository envelopeRepository;
     @Autowired private EventRecordRepository eventRecordRepository;
-    @Autowired private EnvelopeSummaryRepository repo;
+    @Autowired private EnvelopeSummaryRepository envelopeSummaryRepository;
     @Autowired private DbHelper dbHelper;
 
     private static final String CONTAINER_1 = "container1";
@@ -42,11 +37,8 @@ public class EnvelopeSummaryRepositoryTest {
     private static final String BULKSCAN_CONTAINER = "bulkscan";
     private static final String FILE_1_1 = "file_name_1_1";
     private static final String FILE_1_2 = "file_name_1_2";
-    private static final String FILE_1_3 = "file_name_1_3";
     private static final String FILE_2_1 = "file_name_2_1";
     private static final String FILE_2_2 = "file_name_2_2";
-    private static final String FILE_2_3 = "file_name_2_3";
-    private static final String FILE_2_4 = "file_name_2_4";
     private static final String FILE_BULKSCAN_1 = "file_name_bulkscan_1";
 
     @BeforeEach
@@ -61,12 +53,8 @@ public class EnvelopeSummaryRepositoryTest {
         DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
         Instant createdAt1 = LocalDateTime.parse("2019-12-19 10:31:25", formatter).toInstant(UTC);
         Instant createdAt2 = LocalDateTime.parse("2019-12-20 11:32:26", formatter).toInstant(UTC);
-        Instant createdAt3 = LocalDateTime.parse("2019-12-20 12:33:27", formatter).toInstant(UTC);
-        Instant createdAt4 = LocalDateTime.parse("2019-12-20 12:34:28", formatter).toInstant(UTC);
-        Instant createdAt5 = LocalDateTime.parse("2019-12-20 12:35:29", formatter).toInstant(UTC);
-        Instant createdAt6 = LocalDateTime.parse("2019-12-20 12:36:30", formatter).toInstant(UTC);
-        Instant createdAt7 = LocalDateTime.parse("2019-12-20 12:37:31", formatter).toInstant(UTC);
-        Instant createdAt8 = LocalDateTime.parse("2019-12-21 13:38:32", formatter).toInstant(UTC);
+        Instant createdAt3 = LocalDateTime.parse("2019-12-20 12:34:28", formatter).toInstant(UTC);
+        Instant createdAt4 = LocalDateTime.parse("2019-12-21 13:38:32", formatter).toInstant(UTC);
         Instant dispatchedAt = LocalDateTime.parse("2019-12-22 13:39:33", formatter).toInstant(UTC);
 
         // before the request date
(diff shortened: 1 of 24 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
```diff
@@ -0,0 +1,48 @@
...
+        NamedParameterJdbcTemplate jdbcTemplate,
+        EnvelopeSummaryMapper mapper
+    ) {
+        this.jdbcTemplate = jdbcTemplate;
+        this.mapper = mapper;
+    }
+
+    public List<EnvelopeSummary> find(Instant from, Instant to) {
+        return jdbcTemplate.query(
+            "SELECT e.*, ev2.* "
```
doncem: is it not better to select specific columns we need? certainly help validating mapper

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryMapper.java
```diff
@@ -0,0 +1,27 @@
...
+@Component
+public class EnvelopeSummaryMapper implements RowMapper<EnvelopeSummary> {
+
+    @Override
+    public EnvelopeSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
+        return new EnvelopeSummary(
+            rs.getString("container"),
+            rs.getString("file_name"),
+            rs.getTimestamp("file_created_at").toInstant(),
+            rs.getTimestamp("dispatched_at").toInstant(),
```
doncem: `dispatched_at` can be null. current statement will throw an error

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+        // when
+        List<EnvelopeSummary> result = repo.find(
+            LocalDate.parse("2019-12-20").atStartOfDay().toInstant(UTC),
+            LocalDate.parse("2019-12-21").atStartOfDay().toInstant(UTC)
+        );
+
+        // then
+        assertThat(result.stream()
+                       .collect(toList()))
+            .extracting("container", "fileName", "fileCreatedAt", "dispatchedAt", "status", "lastEvent", "eventNotes")
```
doncem: i think we prefer lambda expression here - it's type/compile safe as well

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+            .containsExactlyInAnyOrder(
+                tuple(CONTAINER_1, FILE_1_2, created2, dispatched, Status.DISPATCHED, DISPATCHED, null),
+                tuple(CONTAINER_1, FILE_1_3, created3, dispatched, Status.DISPATCHED, DELETED, null),
+                tuple(CONTAINER_2, FILE_2_1, created4, dispatched, Status.REJECTED, DUPLICATE_REJECTED, "Duplicate"),
+                tuple(CONTAINER_2, FILE_2_2, created5, dispatched, Status.REJECTED, DELETED_FROM_REJECTED, null),
```
doncem: rejected envelopes don't have dispatched timestamp

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
lgonczar: Maybe it would be best to list specific columns and give them aliases, as event and envelope have some overlap when it comes to column names.

lgonczar: Also, we need to filter out envelopes from bulkscan container. This is probably the best place to do it.

src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/EnvelopeSummary.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private final String container;
+    private final String fileName;
+    private final Instant fileCreatedAt;
+    private final Instant dispatchedAt;
+    private final Status status;
+    private final Event lastEvent;
+    private final String eventNotes;
+
+    public EnvelopeSummary(
```
lgonczar: Perhaps last event creation date would be useful?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryMapper.java
```diff
@@ -0,0 +1,31 @@
...
+            rs.getString("container"),
+            rs.getString("file_name"),
+            rs.getTimestamp("file_created_at").toInstant(),
+            rs.getTimestamp("file_dispatched_at") == null
+                ? null : rs.getTimestamp("file_dispatched_at").toInstant(),
+            Status.valueOf(rs.getString("status")),
+            rs.getBoolean("is_deleted"),
+            rs.getString("event") == null ? null : Event.valueOf(rs.getString("event")),
+            rs.getString("notes"),
+            rs.getTimestamp("event_created_at") == null
```
doncem: event creation is never null. notes though are. unless `getString` handles it there's nothing else to do :)

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
```diff
@@ -0,0 +1,53 @@
...
+                + "    FROM events "
+                + "    GROUP BY container, file_name "
+                + ") ev1 "
+                + "ON (e.container = ev1.container AND e.file_name = ev1.file_name) "
+                + "LEFT OUTER JOIN ("
+                + "    SELECT id, event, notes, created_at "
+                + "    FROM events"
+                + ") ev2 "
+                + "ON ev2.id = ev1.max_id "
+                + "WHERE e.container <> '" + BULKSCAN_CONTAINER_NAME + "' "
```
doncem: why not use prepared statement feature instead?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
```diff
@@ -0,0 +1,188 @@
...
+                new EnvelopeSummary(CONTAINER_2,
+                                    FILE_2_3,
+                                    createdAt6,
+                                    dispatchedAt,
+                                    Status.DISPATCHED,
+                                    false,
+                                    null,
+                                    null,
+                                    null
+                )
```
doncem: first argument should be new line as well.

on other note: so many fields in assertion 🙊

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
apereverzin: Have discussed this with Leszek, decided that we should keep using existing approach unless there are clear benefits in using a new approach

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryMapper.java
apereverzin: only if there are events for the envelope, otherwise it is null

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
lgonczar: Open for discussion, though. If there are good benefits, we might want to move to prepared statements from the approach we're currently using.

src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/EventRecord.java
```diff
@@ -26,4 +26,8 @@ public EventRecord(
         this.event = event;
         this.notes = notes;
     }
+
+    public long getId() {
+        return id;
+    }
```
lgonczar: Why do we need this getter?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/EnvelopeSummary.java
```diff
@@ -0,0 +1,38 @@
...
+        String container,
+        String fileName,
+        Instant fileCreatedAt,
+        Instant dispatchedAt,
+        Status status,
+        boolean isDeleted,
+        Event lastEvent,
+        String eventNotes,
+        Instant eventCreatedAt
```
lgonczar: Since the type of last event is called `lastEvent`, maybe `eventNotes` and `eventCreatedAt` should also have `last` prefix?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryMapper.java
lgonczar: I think the envelope is created together with an event, in a single transaction. Not sure the mapper needs to be aware of it, though.

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepository.java
```diff
@@ -86,7 +86,10 @@ public UUID insert(NewEnvelope envelope) {
                 .addValue("fileName", envelope.fileName)
                 .addValue("fileCreatedAt", Timestamp.from(envelope.fileCreatedAt))
                 .addValue("status", envelope.status.name())
-                .addValue("dispatchedAt", Timestamp.from(envelope.dispatchedAt))
+                .addValue(
+                    "dispatchedAt",
+                    envelope.dispatchedAt == null ? null : Timestamp.from(envelope.dispatchedAt)
+                )
```
lgonczar: We should have some tests around it. Perhaps it would be best to extract this bit into a separate PR then?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
```diff
@@ -0,0 +1,193 @@
...
+import static uk.gov.hmcts.reform.blobrouter.data.model.Event.DUPLICATE_REJECTED;
+import static uk.gov.hmcts.reform.blobrouter.data.model.Event.FILE_PROCESSING_STARTED;
+import static uk.gov.hmcts.reform.blobrouter.data.model.Event.REJECTED;
+
+@ActiveProfiles({"integration-test", "db-test"})
+@SpringBootTest
+public class EnvelopeSummaryRepositoryTest {
+    @Autowired private EnvelopeRepository envelopeRepository;
+    @Autowired private EventRecordRepository eventRecordRepository;
+    @Autowired private EnvelopeSummaryRepository repo;
```
lgonczar: Given two other repositories have long names, it would be good for this one to be called `envelopeSummaryRepository` for clarity.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
```diff
@@ -0,0 +1,193 @@
...
+    private static final String CONTAINER_1 = "container1";
+    private static final String CONTAINER_2 = "container2";
+    private static final String BULKSCAN_CONTAINER = "bulkscan";
+    private static final String FILE_1_1 = "file_name_1_1";
+    private static final String FILE_1_2 = "file_name_1_2";
+    private static final String FILE_1_3 = "file_name_1_3";
+    private static final String FILE_2_1 = "file_name_2_1";
+    private static final String FILE_2_2 = "file_name_2_2";
+    private static final String FILE_2_3 = "file_name_2_3";
+    private static final String FILE_2_4 = "file_name_2_4";
```
lgonczar: This is quite a lot of cases. Are they all needed?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
```diff
@@ -0,0 +1,193 @@
...
+    private static final String FILE_BULKSCAN_1 = "file_name_bulkscan_1";
+
+    @BeforeEach
+    void setUp() {
+        dbHelper.deleteAll();
+    }
+
+    @Test
+    @SuppressWarnings("checkstyle:variabledeclarationusagedistance")
+    void should_find_within_date_range() {
```
lgonczar: This test tests quite a lot at once. It would be worth splitting it into multiple tests, each focused on a different thing. For example:
- that the retrieved data maps correctly to fields
- that the search doesn't return records from outside given date range
- that the search doesn't return results for a different container
- that the search doesn't return results for a different file
- event data comes from the last event

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryMapper.java
doncem: well. event i always present. whereas envelope is created at the certain condition. since there's no link, should just trust on events first

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
doncem: out of the box benefit: automatic escape and quoting

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryMapper.java
apereverzin: I suggest bullet proof approach

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public EnvelopeSummaryRepository(
+        NamedParameterJdbcTemplate jdbcTemplate,
+        EnvelopeSummaryMapper mapper
+    ) {
+        this.jdbcTemplate = jdbcTemplate;
+        this.mapper = mapper;
+    }
+
+    public List<EnvelopeSummary> find(Instant from, Instant to) {
```
rkondratowicz: clearer name or javadoc would be nice? which date are `from` and `to` used for?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/EnvelopeSummary.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private final String container;
+    private final String fileName;
+    private final Instant fileCreatedAt;
+    private final Instant dispatchedAt;
+    private final Status status;
+    private final boolean isDeleted;
+    private final Event lastEvent;
+    private final String lastEventNotes;
+    private final Instant lastEventCreatedAt;
```
rkondratowicz: shouldn't those be public?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
rkondratowicz: agree

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
```diff
@@ -0,0 +1,203 @@
...
+                    null,
+                    null,
+                    null
+                )
+            );
+    }
+
+    @Test
+    @SuppressWarnings("checkstyle:variabledeclarationusagedistance")
+    void should_skip_bulkscan_repository() {
```
rkondratowicz: what's tested here?

apereverzin: Envelopes from bulkscan repository should not be included in reports.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
apereverzin: @rkondratowicz @lgonczar 
There are 3 test methods:
- should_find_within_date_range - checks that files before and after given time range are filtered out, also (as well as other methods) it checks that all fields in EventSummary have correct values including container and file name and that event related fields have their values from the last event
- should_skip_bulkscan_repository - checks that the file from 'bulkscan' container is not included in the result
- should_handle_missing_events - checks that in case of no events for a file the result is still successfully fetched; though we expect that every file should have at least 1 event handling this situation (no events for a file) makes the code base bullet proof

I hope this explains details of the test class

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepositoryTest.java
rkondratowicz: you mean 'container'?

apereverzin: @rkondratowicz Yes, renamed the method

rkondratowicz:
Given https://github.com/hmcts/blob-router-service/pull/298 exists, is this PR still relevant?
If not, can it be closed?

apereverzin:
PR-298

PR #285, opened 2020-02-21
Build connection string to reform scan storage in code

lgonczar:
### Change description ###

Build connection string to reform scan storage in code, so that it doesn't need to be manually set in Key Vault.

Rafał's existing PR in cnp-flux-config (https://github.com/hmcts/cnp-flux-config/pull/2201) already matches chart version from this PR and no other changes are needed.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -16,7 +16,6 @@ java:
         - crime-storage-connection-string
         - flyway-password
         - blob-router-POSTGRES-PASS
-        - storage-reform-connection-string
   environment:
     DB_HOST: reform-scan-blob-router-{{ .Values.global.environment }}.postgres.database.azure.com
     DB_PORT: '5432'
@@ -32,6 +31,7 @@ java:
     HANDLE_REJECTED_FILES_CRON: "0/30 * * * * *"
     TASK_SCAN_DELAY: "4000" # in millis
 
+    STORAGE_URL: https://reformscan.{{ .Values.global.environment }}.platform.hmcts.net
     STORAGE_BULKSCAN_URL: https://bulkscan{{ .Values.global.environment }}.blob.core.windows.net
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 30
     STORAGE_BLOB_PUBLIC_KEY: "nonprod_public_key.der"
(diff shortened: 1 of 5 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -32,6 +31,7 @@ java:
     HANDLE_REJECTED_FILES_CRON: "0/30 * * * * *"
     TASK_SCAN_DELAY: "4000" # in millis
 
+    STORAGE_URL: https://reformscan.{{ .Values.global.environment }}.platform.hmcts.net
```
doncem: just double checking, is prod env included in this template?

lgonczar: Prod already overrides this setting: https://github.com/hmcts/cnp-flux-config/blob/master/k8s/prod/common/reform-scan/blob-router.yaml#L27

doncem: cool beans 👍

lgonczar: Deleted my already answered my comment by accident. Bringing it back:
> Prod already overrides this setting: https://github.com/hmcts/cnp-flux-config/blob/master/k8s/prod/common/reform-scan/blob-router.yaml#L27

src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
```diff
@@ -31,9 +31,17 @@ public LeaseClientProvider getLeaseClientProvider() {
...
+        @Value("${storage.account-key}") String accountKey,
+        @Value("${storage.url") String storageUrl
     ) {
+        String connectionString = String.format(
+            "DefaultEndpointsProtocol=https;BlobEndpoint=%s;AccountName=%s;AccountKey=%s",
+            storageUrl,
+            accountName,
+            accountKey
+        );
+
```
mdayican: I think we do not need to create connection string ,we can create the client by directly setting values.

lgonczar: I think that's what Nitin tried initially and it didn't work. Is this correct @nitinprabhu ?

nitinprabhu: > I think we do not need to create connection string ,we can create the client by directly setting values.

what do you mean @mdayican ? example ?

lgonczar: I'm gonna merge this now, as this unblock production deployment. We can change the way we build the client at any point.

PR #286, opened 2020-02-21
access bulk-scan storage via app gateway

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1063

### Change description ###

We can reach the bulk-scan app gateway only by proxy. proxy for bulk_scan introduced.
flux changes:
https://github.com/hmcts/cnp-flux-config/pull/2254

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -94,3 +94,8 @@ scheduling:
       cron: 0 0 6 * * *
     scan:
       delay: ${TASK_SCAN_DELAY} # In milliseconds
+
+
+proxy:
+  host-name: ${PROXY_HOST_NAME:proxyout.reform.hmcts.net}
+  port: ${PROXY_PORT:8080}
```
doncem: perhaps env vars are not needed

src/main/java/uk/gov/hmcts/reform/blobrouter/config/HttpConfiguration.java
```diff
@@ -28,6 +33,23 @@ public HttpComponentsClientHttpRequestFactory clientHttpRequestFactory() {
         return new HttpComponentsClientHttpRequestFactory(getHttpClient());
     }
 
+    @Bean
+    public com.azure.core.http.HttpClient azureHttpClient(
+        @Value("${proxy.host-name}") String proxyHostName,
+        @Value("${proxy.port}") int proxyPort) {
```
doncem: style: closing brace should be in new line ;)

src/main/resources/application.yaml
mdayican: I added for future use if we need in product chart or for other env..

doncem: if need to override it'll be bootstrapped then. giving the current setup of our apps

charts/reform-scan-blob-router/values.yaml
```diff
@@ -33,7 +33,7 @@ java:
     TASK_SCAN_DELAY: "4000" # in millis
 
     STORAGE_URL: https://reformscan.{{ .Values.global.environment }}.platform.hmcts.net
-    STORAGE_BULKSCAN_URL: https://bulkscan{{ .Values.global.environment }}.blob.core.windows.net
+    STORAGE_BULKSCAN_URL: https://bulkscan.{{ .Values.global.environment }}.platform.hmcts.net
```
nitinprabhu: you will need to override this in flux for prod

PR #293, opened 2020-02-25
Fix event recording of process start

doncem:
### Change description ###

There is extra skip/check in blob processor. Moving to container processor instead so that blob processor only deals with actual routing without any skips

```
[ ] Yes
[x] No
```


rkondratowicz:
Or you could just move the `saveEventFileProcessingStarted(...)` call to the correct place.
Not sure we need all changes in this PRs.

Hard for me to see as both refactoring of where 'skipping' happens and moving saving an event are done in single PR.

PR #294, opened 2020-02-25
Move envelope presence check to container processor

doncem:
### Change description ###

Leaving `BlobProcessor`'s responsibility to just process blob after all initial checks are done

```
[ ] Yes
[x] No
```


rkondratowicz:
Can you explain the reasoning behind this change?

doncem:
Minimally explained in description. It is what says: leaving blob processor to start processing (creating that event which was moved #293) by getting blob client, leasing, downloading and so on

PR #296, opened 2020-02-26
Finding the last envelope

apereverzin:

### Change description ###
Finding last envelope by container and file_name after removing unique constraint from envelopes table

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
@@ -182,7 +182,7 @@ public class EnvelopeRepositoryTest {
     }
 
     @Test
-    void should_find_envelopes_by_container_and_file_name() {
+    void should_find_single_envelope_by_container_and_file_name() {
         // given
         final String fileName = "foo.bar";
         final String container = "bar";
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
```diff
@@ -200,6 +200,31 @@ void should_find_envelopes_by_container_and_file_name() {
...
+    @Test
+    void should_find_last_envelope_by_container_and_file_name() {
+        // given
+        final String fileName = "foo.bar";
+        final String container = "bar";
+
+        // and
+        repo.insert(new NewEnvelope(container, fileName, now(), now(), Status.DISPATCHED));
+        // this envelope should be selected
+        UUID envelopeId2 =  repo.insert(
```
rkondratowicz: why 2?

rkondratowicz:
What if we start processing envelope and we get a duplicate. The next job will the use the latest envelope == duplicate

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
apereverzin: The second envelope, there should also be 1 and 3 but they are not used. If you do mind I will rename

apereverzin:
> What if we start processing envelope and we get a duplicate. The next job will the use the latest envelope == duplicate

This PR is just to reflect removal of unique constraint. Currently duplicates are still not allowed, this will be fixed in later PRs

PR #298, opened 2020-02-26
Envelopes report

apereverzin:

### Change description ###
EnvelopeSummary and EnvelopeSummaryRepository, as discussed no event data involved 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
```diff
@@ -0,0 +1,42 @@
...
+        this.jdbcTemplate = jdbcTemplate;
+        this.mapper = mapper;
+    }
+
+    public List<EnvelopeSummary> find(Instant from, Instant to) {
+        return jdbcTemplate.query(
+              "SELECT container, file_name, file_created_at, dispatched_at, "
+                + "       status, is_deleted "
+                + "FROM envelopes "
+                + "WHERE container <> '" + BULKSCAN_CONTAINER_NAME + "' "
```
doncem: so still no to moving to named parameter?

rkondratowicz: Do we event need a new repository with new model and new query.
Can we just read all envelopes and filter out the ones from `bulk_scan` container in service layer??

doncem: what if report queries get nastier? and in real life they should be - complete all in query as it's fast and respond with csv or whatever. would be very wasteful to get data and then filter/amend data in service layer.

since we know it's report, i would vote to rename repository to `ReportRepository` to keep all those in there and each method will represent specific report

apereverzin: Will rename to ReportRepository
And having report related stuff in a separate repo makes it future proof in case of new requirements

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/ReportRepositoryTest.java
```diff
@@ -20,9 +20,9 @@
 
 @ActiveProfiles({"integration-test", "db-test"})
 @SpringBootTest
-public class EnvelopeSummaryRepositoryTest {
+public class ReportRepositoryTest {
```
doncem: perhaps just keep as it was or `EnvelopeSummaryReportTest` - class is big and can be bigger. Let's keep each report tested separately

src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeSummaryRepository.java
rkondratowicz: ok

apereverzin:
PR-305

PR #299, opened 2020-02-26
create new PostgreSQL 11 DB

mdayican:
## JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/BPS-1055

### Change description ###

create new PostgreSQL 11 DB and create keyvaults

```
[ ] Yes
[ x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -149,13 +149,13 @@ resource "azurerm_key_vault_secret" "flyway_password" {
 # endregion
 
 # region NEW DB secrets
-resource "azurerm_key_vault_secret" "REFORM_POSTGRES-USER" {
+resource "azurerm_key_vault_secret" "REFORM_POSTGRES_USER" {
   name         = "${var.component}-REFORM-POSTGRES-USER"
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
   value        = "${module.reform-blob-router-db.user_name}"
 }
 
-resource "azurerm_key_vault_secret" "REFORM_POSTGRES-PASS" {
+resource "azurerm_key_vault_secret" "REFORM_POSTGRES_PASS" {
   name         = "${var.component}-REFORM-POSTGRES-PASS"
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
   value        = "${module.reform-blob-router-db.postgresql_password}"
@@ -180,7 +180,7 @@ resource "azurerm_key_vault_secret" "REFORM_POSTGRES_DATABASE" {
 }
 
 # Copy postgres password for flyway migration
-resource "azurerm_key_vault_secret" "reform_flyway_password" {
+resource "azurerm_key_vault_secret" "REFORM_FLYWAY_PASSWORD" {
   name         = "reform-flyway-password"
   value        = "${module.reform-blob-router-db.postgresql_password}"
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -149,38 +149,38 @@ resource "azurerm_key_vault_secret" "flyway_password" {
 # endregion
 
 # region NEW DB secrets
-resource "azurerm_key_vault_secret" "DB_USER" {
+resource "azurerm_key_vault_secret" "db_user" {
   name         = "${var.component}-db-user"
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
   value        = "${module.reform-blob-router-db.user_name}"
 }
 
-resource "azurerm_key_vault_secret" "DB_PASS" {
+resource "azurerm_key_vault_secret" "db_pass" {
   name         = "${var.component}-db-pass"
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
   value        = "${module.reform-blob-router-db.postgresql_password}"
 }
 
-resource "azurerm_key_vault_secret" "DB_HOST" {
+resource "azurerm_key_vault_secret" "db_host" {
   name         = "${var.component}-db-host"
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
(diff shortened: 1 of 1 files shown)


infrastructure/main.tf
```diff
@@ -99,37 +114,37 @@ resource "azurerm_key_vault_secret" "error_notifications_username" {
 resource "azurerm_key_vault_secret" "POSTGRES-USER" {
   name         = "${var.component}-POSTGRES-USER"
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
-  value        = "${module.blob-router-db.user_name}"
+  value        = "${module.reform-blob-router-db.user_name}"
```
nitinprabhu: should we define new resource ?

mdayican: as it is not live , we can directly start using new one?

doncem: since blob router is not live, can we just bump the version in the original module and that's it? we don't care about any data in any environments atm now do we?

nitinprabhu: will bumping the version destroy and recreate a new DB ?

doncem: i sure hope so

nitinprabhu: ok currently there is a lock on database which needs to be removed

mdayican: tried in perftest, getting error:
Error: Error waiting for deployment: Code="DeploymentFailed" Message="At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details." Details=[{"code":"Conflict","message":"{\r\n  \"status\": \"Failed\",\r\n  \"error\": {\r\n    \"code\": \"ResourceDeploymentFailure\",\r\n    \"message\": \"The resource operation completed with terminal provisioning state 'Failed'.\",\r\n    \"details\": [\r\n      {\r\n        \"code\": \"UnsupportedUpdate\",\r\n        \"message\": \"Cannot update Version for server\"\r\n      }\r\n    ]\r\n  }\r\n}"}]

infrastructure/main.tf
```diff
@@ -21,6 +21,20 @@ module "blob-router-db" {
   subscription       = "${var.subscription}"
 }
 
+module "reform-blob-router-db" {
+  source             = "git@github.com:hmcts/cnp-module-postgres?ref=master"
+  product            = "${var.product}-${var.component}"
+  location           = "${var.location_db}"
+  env                = "${var.env}"
+  database_name      = "reform-blob_router"
```
nitinprabhu: reform_blob_router ?

infrastructure/main.tf
```diff
@@ -133,3 +147,42 @@ resource "azurerm_key_vault_secret" "flyway_password" {
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
 }
 # endregion
+
+# region NEW DB secrets
+resource "azurerm_key_vault_secret" "REFORM_POSTGRES-USER" {
```
aliveni: is this supposed to be REFORM_POSTGRES`_USER`?

infrastructure/main.tf
```diff
@@ -133,3 +147,42 @@ resource "azurerm_key_vault_secret" "flyway_password" {
...
+  name         = "${var.component}-REFORM-POSTGRES-USER"
+  key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
+  value        = "${module.reform-blob-router-db.user_name}"
+}
+
+resource "azurerm_key_vault_secret" "REFORM_POSTGRES-PASS" {
```
aliveni: same here

infrastructure/main.tf
```diff
@@ -21,6 +21,20 @@ module "blob-router-db" {
   subscription       = "${var.subscription}"
 }
 
+module "reform-blob-router-db" {
+  source             = "git@github.com:hmcts/cnp-module-postgres?ref=master"
+  product            = "${var.product}-${var.component}"
+  location           = "${var.location_db}"
+  env                = "${var.env}"
+  database_name      = "reform_blob_router"
```
doncem: should it not be `blob_router`? it's the name of the schema/database

infrastructure/main.tf
```diff
@@ -133,3 +147,42 @@ resource "azurerm_key_vault_secret" "flyway_password" {
   key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
 }
 # endregion
+
+# region NEW DB secrets
+resource "azurerm_key_vault_secret" "REFORM_POSTGRES_USER" {
+  name         = "${var.component}-REFORM-POSTGRES-USER"
```
doncem: can we do smth about those names please? and follow some nice guidance:

- lower case
- resource names with underscores
- vault secret names with `-`

i suggest instead `REFORM-POSTGRES` just use `db` - no need to know what db is it. normally we all know what values are needed for db connection: host, port, username, password, db name

infrastructure/main.tf
mdayican: we already have that name

infrastructure/main.tf
mdayican: these are existing names  just reform prefix added.

infrastructure/main.tf
```diff
@@ -21,6 +21,20 @@ module "blob-router-db" {
   subscription       = "${var.subscription}"
 }
 
+module "reform-blob-router-db" {
+  source             = "git@github.com:hmcts/cnp-module-postgres?ref=master"
+  product            = "${var.product}-${var.component}"
```
doncem: I also doubt this terraform module will allow to create this db as the name will clash. We are re-using product, env so this will produce same Azure Database resource there. Well - will try to but since it is existing it will fail so some other name should be used. I suggest to discuss about all this with the team actually

infrastructure/main.tf
doncem: resource names are still capital case

mdayican: existing ones are capital , shall we create another pr to fix them ?

doncem: existing ones will be removed once migration is complete. they were copied and ignored suggestions

PR #300, opened 2020-02-26
BUGFIX: Set `dispatched_at` to null when creating rejected envelope

rkondratowicz:
Existing test that tested both service methods at the same time was removed and two new tests were added.
Best to review in 'Split' mode.


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -55,17 +58,10 @@ public UUID createRejectedEnvelope(
...
-
-    private UUID createEnvelope(String containerName, String blobName, Instant blobCreationDate, Status status) {
-        return envelopeRepository.insert(new NewEnvelope(
-            containerName,
-            blobName,
-            blobCreationDate,
-            now(),
-            status
-        ));
+        return envelopeRepository
```
nitinprabhu: should we also create event corresponding to this envelope or will that be introduced separately ?

doncem: it is already created one line above

PR #302, opened 2020-02-26
Add service method for creating new envelope

rkondratowicz:

To be used by the BlobProcessor in next PR

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
@@ -39,6 +39,16 @@ public class EnvelopeService {
         return envelopeRepository.find(blobName, containerName);
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -39,6 +39,17 @@ public EnvelopeService(
         return envelopeRepository.find(blobName, containerName);
     }
 
+    public UUID createNewEnvelope(String containerName, String blobName, Instant blobCreationDate) {
```
nitinprabhu: need Transactional annotation on method ?

rkondratowicz: good point

rkondratowicz: fixed

PR #303, opened 2020-02-26
Create an event when error occurs

rkondratowicz:

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
@@ -101,7 +101,7 @@ public class EnvelopeService {
 
     @Transactional
     public void saveEventError(String containerName, String blobName) {
-        eventRecordRepository.insert(new NewEventRecord(containerName, blobName, Event.DUPLICATE_REJECTED));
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -99,6 +99,11 @@ public void saveEventDuplicateRejected(String containerName, String blobName) {
         eventRecordRepository.insert(new NewEventRecord(containerName, blobName, Event.DUPLICATE_REJECTED));
     }
 
+    @Transactional
+    public void saveEventError(String containerName, String blobName) {
```
rkondratowicz: following existing naming convention (see methods above)

nitinprabhu: may be worth passing even name to this method as a parameter. 
```saveEventError(String containerName, String blobName, Event event )```

rkondratowicz: yeah, I don't think we need a separate method for every event.
But I'd have to change that for the other 3 already existing methods.
Let's leave that for a separate refactoring PR.

PR #309, opened 2020-02-27
Report service

apereverzin:

### Change description ###
ReportService along with response classes and unit test

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/DailyReportResponse.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.hmcts.reform.blobrouter.model.out;
+
+import java.time.LocalDate;
+import java.util.List;
+
+public class DailyReportResponse {
+    public final LocalDate date;
+    public final List<EnvelopeSummaryResponse> envelopes;
+    public final int dispatched;
+    public final int rejected;
```
rkondratowicz: was this requested? What do you plan to do with those fields? I think we want to send a .csv file.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
```diff
@@ -0,0 +1,79 @@
...
+        this.reportRepository = reportRepository;
+    }
+
+    public DailyReportResponse getDailyReport(LocalDate date) {
+
+        Instant from = date.atStartOfDay().toInstant(UTC);
+        Instant to = date.atStartOfDay().plusDays(1).toInstant(UTC);
+
+        AtomicInteger rejected = new AtomicInteger(0);
+        AtomicInteger dispatched = new AtomicInteger(0);
```
rkondratowicz: `var` can be used

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
```diff
@@ -0,0 +1,79 @@
...
+                .map(s -> {
+                    if (s.status == DISPATCHED) {
+                        dispatched.getAndIncrement();
+                    } else if (s.status == REJECTED) {
+                        rejected.getAndIncrement();
+                    }
+                    return new EnvelopeSummaryResponse(
+                        s.container,
+                        s.fileName,
+                        ofInstant(s.fileCreatedAt, UTC).toLocalDate(),
```
rkondratowicz: static import hides information here. 'of instant' what?

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/EnvelopeSummaryResponse.java
```diff
@@ -0,0 +1,35 @@
+package uk.gov.hmcts.reform.blobrouter.model.out;
+
+import java.time.LocalDate;
+import java.time.LocalTime;
+
+public class EnvelopeSummaryResponse {
+    public final String container;
+    public final String fileName;
+    public final LocalDate dateReceived;
+    public final LocalTime timeReceived;
```
rkondratowicz: do we need to split it in the API response model?

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/DailyReportResponse.java
apereverzin: David did mention totals in the report. 
In processor we return json (without totals, just list of summaries). For email report DailyReportResponse::envelopes will be converted into csv as it is done in processor

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/EnvelopeSummaryResponse.java
apereverzin: This is how it is done in processor. It makes sense in my opinion

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/DailyReportResponse.java
apereverzin: Will remove

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
apereverzin: Will do

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
apereverzin: Will fix

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/EnvelopeSummaryResponse.java
rkondratowicz: ok

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
```diff
@@ -0,0 +1,59 @@
...
+    public List<EnvelopeSummaryResponse> getDailyReport(LocalDate date) {
+
+        var from = date.atStartOfDay().toInstant(UTC);
+        var to = date.atStartOfDay().plusDays(1).toInstant(UTC);
+
+        return reportRepository.getEnvelopeSummary(from, to).stream()
+                .map(s -> new EnvelopeSummaryResponse(
+                        s.container,
+                        s.fileName,
+                        LocalDateTime.ofInstant(s.fileCreatedAt, UTC).toLocalDate(),
```
rkondratowicz: can the helper method be used here as well?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
```diff
@@ -0,0 +1,59 @@
...
+    private LocalDate toLocalDate(Instant instant) {
+        if (instant != null) {
+            return LocalDateTime.ofInstant(instant, UTC).toLocalDate();
+        }
+        return null;
+    }
+
+    private LocalTime toLocalTime(Instant instant) {
+        if (instant != null) {
+            return LocalTime.parse(DateTimeFormatter.ofPattern(TIME_FORMAT).format(instant.atZone(UTC)));
```
rkondratowicz: do we need to go through a string?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
apereverzin: This is how it is done in processor and the reason is that fileCreatedAt cannot be null

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/EnvelopeSummaryResponse.java
```diff
@@ -0,0 +1,35 @@
+package uk.gov.hmcts.reform.blobrouter.model.out;
+
+import java.time.LocalDate;
+import java.time.LocalTime;
+
+public class EnvelopeSummaryResponse {
```
lgonczar: This class doesn't represent a response, but just an item in the list. Could it be renamed to sth like `EnvelopeSummaryItem`?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
```diff
@@ -0,0 +1,57 @@
...
+    private final ReportRepository reportRepository;
+
+    public ReportService(ReportRepository reportRepository) {
+        this.reportRepository = reportRepository;
+    }
+
+    public List<EnvelopeSummaryResponse> getDailyReport(LocalDate date) {
+
+        var from = date.atStartOfDay().toInstant(UTC);
+        var to = date.atStartOfDay().plusDays(1).toInstant(UTC);
```
lgonczar: We have to use the right time zone for London. Can't be UTC, because it may give inaccurate results.

lgonczar: We use `Europe/London` in the processor

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
```diff
@@ -0,0 +1,57 @@
...
+
+        return reportRepository.getEnvelopeSummary(from, to).stream()
+                .map(s -> new EnvelopeSummaryItem(
+                        s.container,
+                        s.fileName,
+                        toLocalDate(s.fileCreatedAt),
+                        toLocalTime(s.fileCreatedAt),
+                        toLocalDate(s.dispatchedAt),
+                        toLocalTime(s.dispatchedAt),
```
lgonczar: We need to handle null values here

apereverzin: toLocalDate, toLocalTime return null in case of null value

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
apereverzin: Mmm... It is UTC in processor

lgonczar: We'll need to update the processor then, too.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
lgonczar: Aaa, sorry, didn't notice those were this class's methods.

src/test/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+            .willReturn(asList(es1, es2));
+
+        LocalDate dt = LocalDate.of(2019, 1, 14);
+
+        // when
+        List<EnvelopeSummaryItem> res = reportService.getDailyReport(dt);
+
+        // then
+        assertThat(res)
+            .extracting(env -> tuple(
```
lgonczar: Instead of creating the tuple, we could use `.usingFieldByFieldElementComparator()` and compare it with an expected list of envelopes.

src/test/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+    void setUp() {
+        reportService = new ReportService(reportRepository);
+    }
+
+    @Test
+    void getDailyReport_should_convert_date_into_date_range() {
+        // given
+        LocalDate dt = LocalDate.of(2019, 1, 14);
+        Instant from = instant("2019-01-14 00:00:00");
+        Instant to = instant("2019-01-15 00:00:00");
```
lgonczar: It would be best to move these two variables to `then` section and maybe add `expected` prefix.

src/test/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static uk.gov.hmcts.reform.blobrouter.data.model.Status.REJECTED;
+import static uk.gov.hmcts.reform.blobrouter.util.DateTimeUtils.instant;
+import static uk.gov.hmcts.reform.blobrouter.util.DateTimeUtils.localDate;
+import static uk.gov.hmcts.reform.blobrouter.util.DateTimeUtils.localTime;
+
+@ExtendWith(MockitoExtension.class)
+class ReportServiceTest {
+    public static final String CONTAINER_1 = "cont1";
+    public static final String FILE_NAME_1 = "file1.zip";
+    public static final String FILE_NAME_2 = "file2.zip";
```
lgonczar: Since these constants are used in just one test, best to move them to that test.

src/test/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+        verify(reportRepository).getEnvelopeSummary(from, to);
+        verifyNoMoreInteractions(reportRepository);
+    }
+
+    @Test
+    void getDailyReport_should_convert_repo_result() {
+        // given
+        Instant inst1 = instant("2019-01-14 10:11:12");
+        Instant inst2 = instant("2019-01-15 11:12:13");
+        Instant inst3 = instant("2019-01-16 12:13:14");
```
lgonczar: As these instants don't provide any context and are only referenced once, it would be ideal not to define them at all and use their values directly instead.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportService.java
apereverzin: @lgonczar Done. I will also raise a PR to fix this in BlobMover. For the processor I assume we need discussion and a separate Jira task

src/main/java/uk/gov/hmcts/reform/blobrouter/util/TimeZones.java
```diff
@@ -1,8 +1,11 @@
 package uk.gov.hmcts.reform.blobrouter.util;
 
+import java.time.ZoneId;
+
 public final class TimeZones {
 
     public static final String EUROPE_LONDON = "Europe/London";
+    public static final ZoneId EUROPE_LONDON_ZONE_ID = ZoneId.of(EUROPE_LONDON);
```
lgonczar: We'll probably have no more than this one time zone. How about moving it to DateTimeUtils, then? Also, as `EUROPE_LONDON` isn't referenced from anywhere else, probably best to use the value directly when creating zone id.

lgonczar: Sorry, didn't notice this class already existed.

src/test/java/uk/gov/hmcts/reform/blobrouter/services/report/ReportServiceTest.java
```diff
@@ -0,0 +1,101 @@
...
+import static uk.gov.hmcts.reform.blobrouter.data.model.Status.REJECTED;
+import static uk.gov.hmcts.reform.blobrouter.util.DateTimeUtils.instant;
+import static uk.gov.hmcts.reform.blobrouter.util.DateTimeUtils.localDate;
+import static uk.gov.hmcts.reform.blobrouter.util.DateTimeUtils.localTime;
+
+@ExtendWith(MockitoExtension.class)
+class ReportServiceTest {
+    private static final String CONTAINER_1 = "cont1";
+    private static final String FILE_NAME_1 = "file1.zip";
+    private static final String FILE_NAME_2 = "file2.zip";
```
lgonczar: These constants live in the scope of one tests, so they should be defined in it.

src/main/java/uk/gov/hmcts/reform/blobrouter/util/TimeZones.java
apereverzin: Yes, and DateTimeUtils exist only in test modules

PR #310, opened 2020-02-27
Add service method for marking envelope as dispatched

rkondratowicz:
In the new flow we will not create an envelope in `DISPATCHED` status.
Instead, we'll update existing envelope (and create a new event).

Adding a method that will allow this.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/exceptions/EnvelopeNotFoundException.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/exceptions/EnvelopeNotFoundException.java
@@ -1,4 +1,11 @@
 package uk.gov.hmcts.reform.blobrouter.exceptions;
 
 public class EnvelopeNotFoundException extends RuntimeException {
+
+    public EnvelopeNotFoundException() {
+    }
+
+    public EnvelopeNotFoundException(String message) {
+        super(message);
+    }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/exceptions/EnvelopeNotFoundException.java
```diff
@@ -1,4 +1,11 @@
 package uk.gov.hmcts.reform.blobrouter.exceptions;
 
 public class EnvelopeNotFoundException extends RuntimeException {
+
+    public EnvelopeNotFoundException() {
+    }
+
+    public EnvelopeNotFoundException(String message) {
+        super(message);
+    }
```
rkondratowicz: I thought it would be nice to provide some context.
Not changing old code throwing this exception, hence the empty constructor.

rkondratowicz:
> is throwing exception desired behaviour?
Yes, trying to update something that does not exist is not ok.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -87,6 +87,21 @@ public UUID createRejectedEnvelope(
         return envelopeRepository.find(Status.DISPATCHED, containerName, false);
     }
 
+    @Transactional
+    public void markAsDispatched(UUID id) {
+        envelopeRepository
+            .find(id)
+            .ifPresentOrElse(
+                env -> {
+                    envelopeRepository.updateStatus(id, Status.DISPATCHED);
```
mdayican: may be we should consider to create updateStatus  method? at least private.

PR #312, opened 2020-02-27
Add service method for marking envelope as rejected

rkondratowicz:
In the new flow we will not create an envelope in REJECTED status.
Instead, we'll update existing envelope (and create a new event).

Adding a method that will allow this.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/exceptions/EnvelopeNotFoundException.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/exceptions/EnvelopeNotFoundException.java
@@ -1,4 +1,10 @@
 package uk.gov.hmcts.reform.blobrouter.exceptions;
 
 public class EnvelopeNotFoundException extends RuntimeException {
+    public EnvelopeNotFoundException() {
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/exceptions/EnvelopeNotFoundException.java
```diff
@@ -1,4 +1,10 @@
 package uk.gov.hmcts.reform.blobrouter.exceptions;
 
 public class EnvelopeNotFoundException extends RuntimeException {
+    public EnvelopeNotFoundException() {
+    }
+
+    public EnvelopeNotFoundException(String message) {
+        super(message);
```
aliveni: This already exists in #310

rkondratowicz: it's not merged, not available here.

rkondratowicz: whichever PR get's merged first will get that into master, then after refresh this change will disappear

PR #315, opened 2020-02-27
Use new db

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1055

### Change description ###

new db created #299,
now changing the connection configuration .
in next pr will remove obsolete configuration

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -14,17 +14,17 @@ java:
         - storage-account-primary-key
         - storage-account-secondary-key
         - crime-storage-connection-string
-        - flyway-password
-        - blob-router-POSTGRES-PASS
+        - db-flyway-password
+        - blob-router-db-pass
   environment:
-    DB_HOST: reform-scan-blob-router-{{ .Values.global.environment }}.postgres.database.azure.com
+    DB_HOST: blob-router-{{ .Values.global.environment }}.postgres.database.azure.com
     DB_PORT: '5432'
     DB_NAME: blob_router
-    DB_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
+    DB_USER: blob_router@blob-router-{{ .Values.global.environment }}
     DB_CONN_OPTIONS: ?sslmode=require
 
-    FLYWAY_URL: jdbc:postgresql://reform-scan-blob-router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/blob_router?sslmode=require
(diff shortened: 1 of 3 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -14,17 +14,17 @@ java:
...
     DB_PORT: '5432'
     DB_NAME: blob_router
-    DB_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
+    DB_USER: blob_router@blob-router-{{ .Values.global.environment }}
     DB_CONN_OPTIONS: ?sslmode=require
 
-    FLYWAY_URL: jdbc:postgresql://reform-scan-blob-router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/blob_router?sslmode=require
-    FLYWAY_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
+    FLYWAY_URL: jdbc:postgresql://blob-router-{{ .Values.global.environment }}.postgres.database.azure.com:5432/blob_router?sslmode=require
+    FLYWAY_USER: blob_router@blob-router-{{ .Values.global.environment }}
```
doncem: if "fixing" the above then need to clarify how those can be plugged in so the db step in pipeline can execute. spring application can be derived from existing config but not sure about db migration step :thinking:

charts/reform-scan-blob-router/values.yaml
```diff
@@ -14,17 +14,17 @@ java:
...
-        - blob-router-POSTGRES-PASS
+        - db-flyway-password
+        - blob-router-db-pass
   environment:
-    DB_HOST: reform-scan-blob-router-{{ .Values.global.environment }}.postgres.database.azure.com
+    DB_HOST: blob-router-{{ .Values.global.environment }}.postgres.database.azure.com
     DB_PORT: '5432'
     DB_NAME: blob_router
-    DB_USER: blob_router@reform-scan-blob-router-{{ .Values.global.environment }}
+    DB_USER: blob_router@blob-router-{{ .Values.global.environment }}
```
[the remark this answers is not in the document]
mdayican: not now, I do not want mix that with upgrade.

mdayican:
> Can you please use appropriate commit messages ?

this commits happened because github was down.

charts/reform-scan-blob-router/values.yaml
mdayican: I think we need code change,I do not want mix that with upgrade.

mdayican:
as requested

PR #317, opened 2020-02-27
Add test certificate thumbprints in API Mgmt for perftest

lgonczar:
### Change description ###

Configure test certificate thumbprints in API Mgmt for perftest.

This PR is same as #311, which seemed to have permanently broken build.

```
[ ] Yes
[x] No
```


lgonczar:
Merging, as the PR is hanging on the build that has actually passed already.

PR #320, opened 2020-02-28
IGNORE

mdayican:
### Change description ###

get rid of flyway env vars after that we can use all db values from vault

```
[ ] Yes
[ ] No
```


mdayican:
no needed, created other
https://github.com/hmcts/blob-router-service/pull/331
https://github.com/hmcts/blob-router-service/pull/328
https://github.com/hmcts/blob-router-service/pull/326

PR #321, opened 2020-02-28
ReportCsvWriter

apereverzin:

### Change description ###
ReportCsvWriter which will be called while sending reports to email.

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -224,6 +224,8 @@ dependencies {
   implementation group: 'uk.gov.hmcts.reform', name: 'logging-appinsights', version: versions.reformLogging
 
   implementation group: 'uk.gov.hmcts.reform', name: 'properties-volume-spring-boot-starter', version:'0.0.4'
+  implementation group: 'org.apache.commons', name: 'commons-csv', version: '1.8'
+  implementation group: 'org.apache.commons', name: 'commons-collections4', version: '4.1'
```
doncem: is extra dependency really necessary?

apereverzin: it is. in the processor it is transitive via   compile group: 'com.microsoft.azure', name: 'azure-servicebus', version: '1.2.18' but in this case I have decided it should be an overkill

doncem: yea, but is it really necessary? Only used for `CollectionUtils.emptyIfNull`. Can make sure any other way like `Optoinal.ofNullable` for starters. But in general, the incoming `List` should not be empty aka null in first place. Losing the point of being a list

apereverzin: have followed the advice

PR #329, opened 2020-03-02
Create new envelope before processing file

rkondratowicz:
then update it (dispatched/rejected) based on the processing result.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
@@ -70,13 +70,7 @@ public class BlobProcessor {
             leaseClient = leaseClientProvider.get(blobClient);
             leaseClient.acquireLease(60);
 
-            UUID id =
-                envelopeService
-                    .createNewEnvelope(
-                        containerName,
-                        blobName,
-                        blobCreationDate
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -60,10 +60,17 @@ public BlobProcessor(
...
         String blobName = blobClient.getBlobName();
         var containerName = blobClient.getContainerName();
+        Instant blobCreationDate = blobClient.getProperties().getLastModified().toInstant();
 
         logger.info("Processing {} from {} container", blobName, containerName);
 
-        envelopeService.saveEvent(containerName, blobName, Event.FILE_PROCESSING_STARTED);
+        UUID id =
+            envelopeService
+                .createNewEnvelope(
```
lgonczar: Should we have an audit event when creating a new envelope?

lgonczar: Also, this should probably moved after the lease is acquired.

rkondratowicz: `. createNewEnvelope()` service method creates an event

PR #330, opened 2020-03-02
Repository changes to get envelopes count in Cft and Crime containers

aliveni:

### Change description ###
Envelope Repository changes to check if there are envelopes with file creation date for the given date-time range.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
@@ -10,6 +10,7 @@ import uk.gov.hmcts.reform.blobrouter.data.model.NewEnvelope;
 import uk.gov.hmcts.reform.blobrouter.data.model.Status;
 
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepository.java
```diff
@@ -125,4 +125,28 @@ public int markAsDeleted(UUID id) {
             new MapSqlParameterSource("id", id)
         );
     }
+
+    public Integer getEnvelopesCountInCftContainers(Instant fromDateTime, Instant toDateTime) {
```
nitinprabhu: can we not have a parameter for container name and then one method can be used to retrieve both type of envelopes ?

aliveni: but Cft containers are more than 1 container

nitinprabhu: Param can be crime or cft ?

aliveni: I added containers parameter which is a collection of containers. For crime, it is just `crime`, and for cft  - it will be `ServiceConfiguration.getSourceConatiners() minus crime`. 
Is this better?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
```diff
@@ -263,13 +265,40 @@ void should_find_envelopes_by_status() {
         assertThat(repo.find(Status.REJECTED, true)).extracting(env -> env.fileName).containsExactly("f4", "f8");
     }
 
+    @Test
+    void should_return_envelopes_count_for_container_and_time_range() {
+        //given
+        Instant fromDate = now();
+        Instant toDate = fromDate.plus(3, ChronoUnit.MINUTES);
```
rkondratowicz: A bit weird usage.
Shall we search from the past til now instead?

aliveni: Updated the test

PR #331, opened 2020-03-02
Get rid off flyway variables

mdayican:
### Change description ###

get rid of flyway variables. instead use db variables 
next pr will try to use db vars from vault.

```
[ ] Yes
[x] No
```


src/integrationTest/resources/application.properties
```diff
@@ -14,6 +14,7 @@ flyway.skip-migrations=false
 
 spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver
 spring.datasource.url=jdbc:tc:postgresql:9.6://localhost/blob_router
+spring.flyway.url=jdbc:tc:postgresql:9.6://localhost/blob_router
```
doncem: we're now on 11, shall we reflect in testcontainers as well?

PR #332, opened 2020-03-02
Daily report properties

apereverzin:

### Change description ###
Charts for sending daily reports

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.yaml
```diff
@@ -10,6 +10,9 @@ java:
         - error-notifications-password
         - error-notifications-url
         - error-notifications-username
+        - reports-email-username
+        - reports-email-password
+        - reports-recipients
```
lgonczar: This one doesn't exist in bootstrap.yaml

charts/reform-scan-blob-router/values.aat.template.yaml
```diff
@@ -1,6 +1,7 @@
 java:
   environment:
     CRIME_ENABLED: "true"
+    SEND_DAILY_REPORT_ENABLED: "false"
```
doncem: should be default in `values.yml` instead

charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -45,6 +45,7 @@ java:
     BULK_SCAN_PROCESSOR_URL: http://bulk-scan-processor-aat.service.core-compute-aat.internal
     CRIME_DESTINATION_CONTAINER: bs-sit-scans-received
     CRIME_ENABLED: true
+    SEND_DAILY_REPORT_ENABLED: false
```
doncem: once in `values.yml` this is not needed either

charts/reform-scan-blob-router/values.yaml
doncem: also missing in perftest and prod environments

doncem: bootstrap is not necessary when doing chart updates

apereverzin:
PR-356

PR #335, opened 2020-03-03
Bug: Don't create events when blob lease cannot be acquired

rkondratowicz:
And don't log an error, just skip that file.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
@@ -74,9 +74,6 @@ public class BlobProcessor {
         envelopeService.saveEvent(containerName, blobName, Event.FILE_PROCESSING_STARTED);
 
         try {
-            leaseClient = leaseClientProvider.get(blobClient);
-
-            leaseClient.acquireLease(60);
             byte[] rawBlob = tryToDownloadBlob(blobClient);
 
             var verificationResult = blobVerifier.verifyZip(blobName, rawBlob);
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -54,8 +56,8 @@
             )
         );
 
-    @Mock BlobClient blobClient;
-    @Mock BlobProperties blobProperties;
+    @Mock(lenient = true) BlobClient blobClient;
+    @Mock(lenient = true) BlobProperties blobProperties;
```
rkondratowicz: `blobExists` method mock all the properties of the blob. Not all are used before the lease is acquired

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessorTest.java
```diff
@@ -220,6 +222,24 @@ void should_not_upload_to_crime_when_blob_does_not_contain_envelope_entry() {
         verify(envelopeService).saveEvent(SOURCE_CONTAINER, "envelope1.zip", Event.FILE_PROCESSING_STARTED);
     }
 
+    @Test
+    void should_not_create_events_when_lease_cant_be_acquired() {
```
nitinprabhu: should_not_create_events_when_lease_cannot_be_acquired ?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -63,14 +63,17 @@ public void process(BlobClient blobClient) {
 
         logger.info("Processing {} from {} container", blobName, containerName);
 
-        envelopeService.saveEvent(containerName, blobName, Event.FILE_PROCESSING_STARTED);
+        BlobLeaseClient leaseClient = leaseClientProvider.get(blobClient);
+        try {
+            leaseClient.acquireLease(60);
+        } catch (Exception exc) {
```
lgonczar: Maybe we should catch some specific exception, that represents a conflict or sth? We still might have problems connecting and that should result in logging an error.

rkondratowicz: done

PR #337, opened 2020-03-04
Rename Event enum to EventType

rkondratowicz:
A proposal


src/main/java/uk/gov/hmcts/reform/blobrouter/data/model/EventType.java
```diff
@@ -1,6 +1,6 @@
 package uk.gov.hmcts.reform.blobrouter.data.model;
 
-public enum Event {
+public enum EventType {
```
apereverzin: For me more sense has EnvelopeEventType or even EnvelopeLifecycleEventType. Or we should rename EventRecord to EnvelopeEvent and this enum to EventType as proposed

PR #339, opened 2020-03-04
Add config for checking new envelopes in containers

aliveni:

### Change description ###
Add config for 
- `cron expression` for new envelopes check
- `TTL` which will be used to query new in the database.

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.yaml
```diff
@@ -27,6 +27,8 @@ java:
     DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
     HANDLE_REJECTED_FILES_CRON: "0/30 * * * * *"
     REJECT_DUPLICATES_CRON: "0/30 * * * * *"
+    CHECK_NEW_ENVELOPES_CRON: "0 55 0 * * MON-FRI" # every hour at 55 minutes on weekdays
```
aliveni: The alert (for checking the logs) will run at every hour so the envelope task can run 5 minutes before so it will be reflected in the app insights. shall I change it to run on every day?

charts/reform-scan-blob-router/values.yaml
```diff
@@ -27,6 +27,8 @@ java:
...
+    CHECK_NEW_ENVELOPES_TTL: PT1H
```
rkondratowicz: what's that?

charts/reform-scan-blob-router/values.yaml
rkondratowicz: we want to check only during business hours I think

charts/reform-scan-blob-router/values.yaml
aliveni: This will be used here: https://github.com/hmcts/blob-router-service/blob/20fb4707aef03ea11ed5ad61ff3e53b59fb7ac4f/src/main/java/uk/gov/hmcts/reform/blobrouter/services/NewEnvelopesFinder.java#L50
If we can always check for the last 1 hour, this config can be removed

aliveni: This value remains the same in all environments, so I'll rename this add this to `aplication.yaml`.

aliveni:
Closing this PR as the build is failing

PR #340, opened 2020-03-04
Create new events table

rkondratowicz:
Now that envelopes are created at the very start of processing, events will have a foreign key to envelope. `file_name` and `container` columns will not be needed.

In this PR a new table is created along with new models and repository.
It will be used in the next PR.
After that, old table (along with models and repositories) will be deleted.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/EnvelopeEvent.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/EnvelopeEvent.java
@@ -1,0 +1,23 @@
+package uk.gov.hmcts.reform.blobrouter.data.eventsv2;
+
+import uk.gov.hmcts.reform.blobrouter.data.events.EventType;
+
+import java.time.Instant;
+import java.util.UUID;
+
+public class EnvelopeEvent {
+
+    public final long id;
+    public final UUID envelopeId;
+    public final EventType type;
+    public final String notes;
+    public final Instant createdAt;
+
+    public EnvelopeEvent(long id, UUID envelopeId, EventType type, String notes, Instant createdAt) {
+        this.id = id;
+        this.envelopeId = envelopeId;
+        this.type = type;
+        this.notes = notes;
+        this.createdAt = createdAt;
+    }
+}
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/NewEnvelopeEvent.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/NewEnvelopeEvent.java
@@ -1,0 +1,18 @@
+package uk.gov.hmcts.reform.blobrouter.data.eventsv2;
+
+import uk.gov.hmcts.reform.blobrouter.data.events.EventType;
+
+import java.util.UUID;
+
+public class NewEnvelopeEvent {
+
+    public final UUID envelopeId;
+    public final EventType type;
+    public final String notes;
+
+    public NewEnvelopeEvent(UUID envelopeId, EventType type, String notes) {
+        this.envelopeId = envelopeId;
+        this.type = type;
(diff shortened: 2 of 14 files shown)


src/main/resources/db/migration/V006__Add_envelope_events_table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE envelope_events
+(
+    id          BIGSERIAL    PRIMARY KEY,
+    envelope_id UUID         NOT NULL REFERENCES envelopes ON DELETE CASCADE,
+    created_at  TIMESTAMP    NOT NULL,
+    type        VARCHAR(100) NOT NULL,
+    notes       TEXT         NULL
```
nitinprabhu: worth renaming this to error_description or something else

src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/NewEnvelopeEvent.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.hmcts.reform.blobrouter.data.eventsv2;
+
+import uk.gov.hmcts.reform.blobrouter.data.events.EventType;
+
+import java.util.UUID;
+
+public class NewEnvelopeEvent {
+
+    public final UUID envelopeId;
+    public final EventType type;
```
doncem: Where is the `EventType`?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/EnvelopeEvent.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.hmcts.reform.blobrouter.data.eventsv2;
```
doncem: personally not very strong on such package

rkondratowicz: I'm open to  alternative names

src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/NewEnvelopeEvent.java
rkondratowicz: Event was renamed to this https://github.com/hmcts/blob-router-service/pull/337

src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/EnvelopeEvent.java
doncem: events can be enough? especially when former models etc will be removed anyway - there's no need to version now then. makes sense?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/NewEnvelopeEvent.java
doncem: my bad, only now pulled latest master - i was checking for changes locally

src/main/java/uk/gov/hmcts/reform/blobrouter/data/eventsv2/EnvelopeEvent.java
rkondratowicz: there's already a package called 'events'

rkondratowicz: oh you mean to put it in the existing one

src/main/resources/db/migration/V006__Add_envelope_events_table.sql
rkondratowicz: Let's stick to existing name from previous model, in case we want to put any notes on successful events.
We can rename it later.

PR #343, opened 2020-03-05
Add config for checking new envelopes in containers

aliveni:

### Change description ###
Duplicate of #339
Add config for new envelopes check cron expression.
Flux PR https://github.com/hmcts/cnp-flux-config/pull/2358

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.yaml
```diff
@@ -27,6 +27,7 @@ java:
     DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
     HANDLE_REJECTED_FILES_CRON: "0/30 * * * * *"
     REJECT_DUPLICATES_CRON: "0/30 * * * * *"
+    CHECK_NEW_ENVELOPES_CRON: "0 55 0 * * MON-FRI" # every hour at 55 minutes on weekdays
```
doncem: it looks like once per day. also missing `?` on a day declaration rather than `*`

apereverzin: Yes, looks like at 0:55 - once per day

aliveni:
PR build is failing. Will create a new PR

PR #345, opened 2020-03-05
Send reports

apereverzin:

### Change description ###
SendDailyReportTask implementing scheduled job to send daily reports.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTaskTest.java
```diff
@@ -0,0 +1,141 @@
...
+        verify(reportService).getDailyReport(any(LocalDate.class));
+        verify(reportCsvWriter).writeEnvelopesSummaryToCsv(report);
+        verifyNoInteractions(emailSender);
+    }
+
+    @Test
+    void should_throw_if_empty_recipients() {
+        assertThrows(
+            RuntimeException.class,
+            () -> sendDailyReportTask = new SendDailyReportTask(
```
rkondratowicz: no need to assign to variable

apereverzin: will fix

apereverzin:
Build is failing, new PR:
https://github.com/hmcts/blob-router-service/pull/352

infrastructure/main.tf
```diff
@@ -126,3 +126,44 @@ resource "azurerm_key_vault_secret" "POSTGRES_DATABASE" {
...
+
+data "azurerm_key_vault_secret" "bulk_scan_reports_email_password" {
+  key_vault_id = "${data.azurerm_key_vault.bulk_scan_key_vault.id}"
+  name         = "reports-email-password"
+}
+
+data "azurerm_key_vault_secret" "bulk_scan_reports_recipients" {
+  key_vault_id = "${data.azurerm_key_vault.bulk_scan_key_vault.id}"
+  name         = "reports-recipients"
+}
```
lgonczar: Should we reuse the exact same secrets as bulk-scan? Even if the values are currently the same, should the secrets be the same?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTask.java
```diff
@@ -0,0 +1,92 @@
...
+        @Value("${spring.mail.username}") String from,
+        @Value("${reports.recipients}") String[] recipients
+    ) {
+        this.reportService = reportService;
+        this.reportCsvWriter = reportCsvWriter;
+        this.emailSender = emailSender;
+        this.from = from;
+
+        if (recipients == null) {
+            this.recipients = new String[0];
```
lgonczar: we can throw the exception straight away

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTask.java
```diff
@@ -0,0 +1,92 @@
...
+            emailSender.sendMessageWithAttachments(
+                EMAIL_SUBJECT,
+                EMAIL_BODY,
+                from,
+                recipients,
+                Map.of(getReportAttachmentName(reportDate), reportFile)
+            );
+        } catch (Exception ex) {
+            logger.error(
+                "Error sending daily report: {}",
```
lgonczar: A very minor thing, but worth saying `for date`, so that noone takes the date as an actual error message. It's a suggestion, up to you.

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTaskTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Captor
+    private ArgumentCaptor<String[]> recipientsCaptor;
+
+    @Captor
+    private ArgumentCaptor<Map<String, File>> attachmentsCaptor;
+
+    @BeforeEach
+    void setUp() {
+        sendDailyReportTask = new SendDailyReportTask(
```
lgonczar: A suggestion: instead of creating an instance of this task in the `setUp` method, we could just call a method creating the task from test. That method, say `newSendDailyReportTask(...)` could take `from` and `recipients` as parameters. This way we could get rid of those two class-global fields and put them where they're used.

infrastructure/main.tf
apereverzin: @lgonczar Yes, checked also with @doncem

apereverzin:
PR-354

PR #351, opened 2020-03-06
Add config for new envelopes check

aliveni:

### Change description ###
Duplicate of #339
Add config for new envelopes check cron expression.
Flux PR https://github.com/hmcts/cnp-flux-config/pull/2358

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.yaml
```diff
@@ -27,6 +27,7 @@ java:
     DELETE_DISPATCHED_FILES_CRON: "0 0/10 * * * *"
     HANDLE_REJECTED_FILES_CRON: "0/30 * * * * *"
     REJECT_DUPLICATES_CRON: "0/30 * * * * *"
+    CHECK_NEW_ENVELOPES_CRON: "0 55 9-17 ? * MON-FRI" # every hour at 55 minutes on weekdays between 9am-6pm
```
nitinprabhu: it's 9 and 5 comment needs update

aliveni: the last run will be at 17:55pm

doncem: it's between 9 and 6 ;)

PR #358, opened 2020-03-09
Make sure appropriate information is logged when when blob download fails

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1043

### Change description ###

Make sure appropriate information is logged when when blob download fails

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
@@ -29,12 +29,15 @@ import java.util.zip.ZipInputStream;
 
 import static com.google.common.io.ByteStreams.toByteArray;
 import static org.slf4j.LoggerFactory.getLogger;
+import static org.springframework.http.HttpStatus.BAD_GATEWAY;
 
 @Component
 @EnableConfigurationProperties(ServiceConfiguration.class)
 public class BlobProcessor {
 
     private static final Logger logger = getLogger(BlobProcessor.class);
+
+    private static final String MESSAGE_FAILED_TO_DOWNLOAD_BLOB = "Failed to download blob";
 
     private final BlobDispatcher dispatcher;
     private final EnvelopeService envelopeService;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -146,11 +147,17 @@ public void process(BlobClient blobClient) {
...
     }
 
-    private byte[] tryToDownloadBlob(BlobClient blobClient) throws IOException {
+    private byte[] downloadBlob(BlobClient blobClient) throws IOException {
         try (var outputStream = new ByteArrayOutputStream()) {
             blobClient.download(outputStream);
 
             return outputStream.toByteArray();
+        } catch (Exception exc) {
+            String errorMessage = exc instanceof IOException
```
lgonczar: We could extend this condition by checking if the cause of this `IOException` is `BlobStorageException` with status code 502. Not sure we should go that deep. Thoughts?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -146,11 +147,17 @@ public void process(BlobClient blobClient) {
         }
     }
 
-    private byte[] tryToDownloadBlob(BlobClient blobClient) throws IOException {
+    private byte[] downloadBlob(BlobClient blobClient) throws IOException {
         try (var outputStream = new ByteArrayOutputStream()) {
             blobClient.download(outputStream);
 
             return outputStream.toByteArray();
+        } catch (Exception exc) {
```
lgonczar: This change doesn't make any difference in this class's logic - only intermediate exception is thrown, still to be caught in this class. No way to unit test this bit. Alternatively, I could extract this method into a separate dependency, but that's a bit much for such a trivial task.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
nitinprabhu: can we not add a catch block for IOException ?

lgonczar: If the condition is as simple as it currently is, sure. I initially made it more complex and it might end up like that.

nitinprabhu: I think checking the contents of the exception message to see if it contains 502(or some string we which get back in the response when files are blocked by AV) that would be nicer

lgonczar: Made some changes here, as I used logs from bulk-scan-processor, which uses an older version of the lib. Catching `BlobStorageException` now, after verifying the behaviour in blob-router.

lgonczar: Made some changes here, as I used logs from bulk-scan-processor, which uses an older version of the lib. Catching `BlobStorageException` now, after verifying the behaviour in blob-router.

lgonczar:
The added tests are not essential, but necessary for the build to pass through quality gate.

PR #361, opened 2020-03-09
Add config for time-interval used to check new envelopes

aliveni:

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -96,6 +96,7 @@ scheduling:
       ttl: PT72H # ISO-8601
       cron: 0 0 6 * * *
     check-new-envelopes:
+      time-interval: 60 # In minutes
```
rkondratowicz: did you consider using `PT60M` instead? We use iso-8601 in link 96 and 83

aliveni: I tried `PT60M` but when trying to get the value, `timeInterval.get(ChronoUnit.MINUTES)`  returns ` java.time.temporal.UnsupportedTemporalTypeException: Unsupported unit: Minutes`.

aliveni: `Duration.get()` supports only a few types. I've updated to use `seconds`.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/NewEnvelopesFinder.java
```diff
@@ -49,7 +56,7 @@ public void checkNewCrimeEnvelopesCreated() {
 
     private void checkNewEnvelopesInContainers(Set<String> containers, String containersGroupName) {
         Instant toDateTime = Instant.now();
-        Instant fromDateTime = toDateTime.minus(1, HOURS); //TODO: read duration from config
+        Instant fromDateTime = toDateTime.minusSeconds(timeInterval.getSeconds());
```
rkondratowicz: would
```java
Instant fromDateTime = toDateTime.minus(timeInterval);
```
work?

aliveni: Yes it works. updated now

PR #362, opened 2020-03-09
Return envelopes from `DuplicateFinder`

rkondratowicz:
Changing the return type
from `BlobItem` (which contains 'name' and 'container')
to `Envelope` (which contains above + envelope ID)

This is in preparation to **new events flow**, where event has a foreign key to envelope.
The ID of envelope will be needed to create an event.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/DuplicateFinder.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/DuplicateFinder.java
@@ -31,7 +31,7 @@ public class DuplicateFinder {
             .stream()
             .map(blob -> envelopeService.findEnvelope(blob.getName(), containerName))
             .flatMap(Optional::stream)
-            .filter(envelope -> envelope.isDeleted)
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/DuplicateFinder.java
```diff
@@ -23,19 +24,14 @@ public DuplicateFinder(
...
 
-    public List<BlobItem> findIn(String containerName) {
+    public List<Envelope> findIn(String containerName) {
         return storageClient
             .getBlobContainerClient(containerName)
             .listBlobs()
             .stream()
-            .filter(blob -> hasAlreadyBeenProcessed(blob.getName(), containerName))
+            .map(blob -> envelopeService.findEnvelope(blob.getName(), containerName))
+            .flatMap(Optional::stream)
```
rkondratowicz: Getting rid of empty optionals (new way from Java 9)
https://www.baeldung.com/java-9-optional#the-stream-method)

PR #368, opened 2020-03-10
Update New Envelopes check scheduled job cron to run every hour

aliveni:

```
[ ] Yes
[x] No
```


aliveni:
Flux PR https://github.com/hmcts/cnp-flux-config/pull/2417

charts/reform-scan-blob-router/values.yaml
```diff
@@ -32,7 +32,7 @@ java:
     REJECT_DUPLICATES_CRON: "0/30 * * * * *"
     CHECK_NEW_ENVELOPES_CRON: "0 55 9-17 ? * MON-FRI" # every hour at 55 minutes on weekdays between 9am-6pm
     TASK_SCAN_DELAY: "4000" # in millis
-    SEND_DAILY_REPORT_CRON: "0 0 18 ? * MON-FRI"
+    SEND_DAILY_REPORT_CRON: "0 0 0/1 * * *"
```
rkondratowicz: same as `"0 0 * * * *"` I think

lgonczar: It is the same

PR #369, opened 2020-03-10
Create error events

rkondratowicz:
When processing a blob.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
@@ -9,6 +9,7 @@ import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.blobrouter.config.ServiceConfiguration;
 import uk.gov.hmcts.reform.blobrouter.config.StorageConfigItem;
 import uk.gov.hmcts.reform.blobrouter.config.TargetStorageAccount;
+import uk.gov.hmcts.reform.blobrouter.data.events.EventType;
 import uk.gov.hmcts.reform.blobrouter.exceptions.InvalidZipArchiveException;
 import uk.gov.hmcts.reform.blobrouter.services.BlobVerifier;
 import uk.gov.hmcts.reform.blobrouter.services.EnvelopeService;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -170,8 +170,14 @@ private void tryToReleaseLease(BlobLeaseClient leaseClient, String blobName, Str
         }
     }
 
-    private void handleError(Exception exc, BlobClient blob) {
-        logger.error("Error occurred while processing {} from {}", blob.getBlobName(), blob.getContainerName(), exc);
-        // TODO: save error event
+    private void handleError(UUID envelopeId, BlobClient blob, Exception exc) {
+        logger.error(
+            "Error occurred while processing blob. File name: {}, Container: {}, Envelope ID: {}",
```
lgonczar: 🥇

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -67,9 +68,8 @@ public void process(BlobClient blobClient) {
             .acquireFor(blobClient)
             .ifPresentOrElse(
                 leaseClient -> {
+                    UUID id = envelopeService.createNewEnvelope(containerName, blobName, blobCreationDate);
                     try {
-                        UUID id = envelopeService.createNewEnvelope(containerName, blobName, blobCreationDate);
-
```
lgonczar: Why was this line move outside the `try` part?

rkondratowicz: so that we have the envelope ID. without it we can't create events in the new flow (foreign key instead of file name + container name).

lgonczar: I guess a failure to create a new envelope is one of the reasons to create an error event. Maybe we should define the id (as null) before the try, create the envelope inside `try` and be prepared that it's null in the `catch` clause?

lgonczar: Aaa, sorry, it's the foreign key.

lgonczar:
> and next if we could try with resource lease client so the `tryToReleaseLease` can be moved to custom autoclosable wrapper.. 🙃

definitely some wrapper for leasing a blob would be nice

PR #371, opened 2020-03-10
remove from cache added

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1053

### Change description ###

remove from cache added.
later we will remove sas token from cache if we get any 400 errors.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BulkScanSasTokenCache.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BulkScanSasTokenCache.java
@@ -45,6 +45,10 @@ public class BulkScanSasTokenCache {
         return tokenCache.get(containerName, c -> this.createSasToken(c));
     }
 
+    public void removeFromCache(String containerName) {
+        tokenCache.invalidate(containerName);
+    }
+
     private String createSasToken(String containerName) {
         return bulkScanSasTokenClient.getSasToken(containerName).sasToken;
     }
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/hmcts/reform/blobrouter/services/storage/BulkScanSasTokenCacheTest.java
```diff
@@ -119,4 +119,34 @@ void should_create_new_sas_token_when_cache_is_expired() {
         verify(bulkScanProcessorClient,times(2)).getSasToken(containerName);
     }
 
+    @Test
+    void should_remove_from_cache_when_cache_invalidated() {
```
nitinprabhu: should_remove_sas_token_from_cache_when_it_is_invalidated

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BulkScanSasTokenCache.java
```diff
@@ -45,6 +45,10 @@ public String getSasToken(String containerName) {
         return tokenCache.get(containerName, c -> this.createSasToken(c));
     }
 
+    public void removeFromCache(String containerName) {
+        tokenCache.invalidate(containerName);
+    }
```
lgonczar: What will call this method and when?

mdayican: later we will remove sas token from cache if we get any 400 errors.

lgonczar: I'm asking because I wonder if it would be better to use the existing method that gets the SAS token and add a parameter that decides if the cache item should be recreated.

mdayican: we are not going to recreate if we got error.

PR #372, opened 2020-03-10
Improved logs

apereverzin:

### Change description ###
Improved logging a bit.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTask.java
@@ -63,6 +63,7 @@ public class SendDailyReportTask {
     @Scheduled(cron = "${scheduling.task.send-daily-report.cron}")
     @SchedulerLock(name = TASK_NAME)
     public void sendReport() {
+        logger.info("Started {} job", TASK_NAME);
         final LocalDate reportDate = LocalDate.now();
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTask.java
```diff
@@ -77,9 +78,10 @@ public void sendReport() {
                 recipients,
                 Map.of(getReportAttachmentName(reportDate), reportFile)
             );
+            logger.info("Finished {} job", TASK_NAME);
```
rkondratowicz: should this not be after catch?

apereverzin: Do not  think so - this indicates success

nitinprabhu: Those logs are about start and finish and not to indicate success ? So if you have an exception you would never see finished ?

apereverzin: Ok, will update @nitinprabhu

doncem: success/failure for these tasks should be indicated with request/dependency feature by app insights. if ever this functionality comes in. on the other hand it can replace querying logs for some alerts. which is absurd

PR #373, opened 2020-03-10
Improved test

apereverzin:

### Change description ###
Minor improvement in test class

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTaskTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTaskTest.java
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.reform.blobrouter.tasks;
 
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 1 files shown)


lgonczar:
Nice. Now that we're creating the task from tests, it would be good to move `FROM` and `recipients` global fields to the method(s) where they're used.

apereverzin:
FROM and recipients are used in several tests and in captors so it makes sense to define them as class attributes not in methods

lgonczar:
> FROM and recipients are used in several tests and in captors so it makes sense to define them as class attributes not in methods

They're only used at test level, so they can definitely be moved to tests.

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/SendDailyReportTaskTest.java
```diff
@@ -113,6 +103,12 @@ void sendReport_should_call_email_sender() throws Exception {
     @Test
     void sendReport_should_not_call_email_sender_if_csv_writer_throws() throws Exception {
         // given
+        String from = "From";
+
+        String[] recipients = new String[]{"rec1", "rec2"};
+
+        sendDailyReportTask = getSendDailyReportTask(from, recipients);
```
lgonczar: These two can even just be passed as values, but that's only a tiny suggestion.

PR #378, opened 2020-03-11
move storage builder to config

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1053

### Change description ###

to increase the testability storage client builder moved to config and it is injectable to provider.
next PR I will change provider and dispatcher and move upload function to provider.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.blobrouter.config;
 
+import com.azure.core.http.HttpClient;
 import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.BlobContainerClientBuilder;
 import com.azure.storage.blob.BlobServiceClient;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/config/StorageConfiguration.java
```diff
@@ -57,4 +58,14 @@ public static BlobContainerClient getCrimeStorageClient(
...
+
+    @Bean("bulk-scan-blob-client-builder")
+    public BlobContainerClientBuilder getBulkScanBlobContainerClientBuilder(
+        HttpClient httpClient,
+        @Value("${storage.bulkscan.url}") String bulkScanStorageUrl
+    ) {
+        return new BlobContainerClientBuilder()
+            .httpClient(httpClient)
+            .endpoint(bulkScanStorageUrl);
+    }
```
lgonczar: Is this builder immutable, i.e. will there be no side effects if it's reused?

mdayican: it is thread safe. 
we still create client for each request.

lgonczar: We execute a sequence of methods on this builder:
```
                    .sasToken(bulkScanSasTokenCache.getSasToken(containerName))
                    .endpoint(bulkScanStorageUrl)
                    .containerName(containerName)
                    .buildClient();
```

Is it not impossible that these calls will be mixed by two different threads so that when each calls `.buildClient()` they might end up with a setup that they didn't expect?

PR #381, opened 2020-03-11
[DO NOT MERGE] Used factory method with user email for green mail user

apereverzin:

### Change description ###
Integration test is failing inconsistently (and very unfrequently) with this error message:
16:01:33  EmailSenderTest > should_send_email_to_all_recepients() FAILED
16:01:33      org.opentest4j.AssertionFailedError: 
16:01:33      Expecting:
16:01:33       <"test_username">
16:01:33      to be equal to:
16:01:33       <"a@b.c">
16:01:33      but was not.

It looks like GreenMail (which is used for integration tests for sending email) is mixing username and 'from' header. In existing code GreenMailUser is created using factory method with username and password, internally GreenMail uses username as user email, and there is a suspicion that this value is also used for 'from' header instead of the value supposed to be used when email is actually sent (submitted as 'from' parameter of EmailSender::sendMessageWithAttachments method).

In this PR factory method using 3 parameters (user email, username, password) is used; it is expected that in this case when user email (which is the same 'a@b.c' as used later when calling EmailSender::sendMessageWithAttachments) is explicitly submitted to GreenMail username will never be used in 'from' header and the failure will go.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/email/EmailSenderTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/services/email/EmailSenderTest.java
@@ -63,7 +63,6 @@ public class EmailSenderTest {
             Map.of(FILE_NAME_1, file1, FILE_NAME_2, file2)
         );
 
-        assertThat(greenMail.getReceivedMessages()).hasSize(RECIPIENTS.length);
         MimeMessageParser msg = new MimeMessageParser(greenMail.getReceivedMessages()[0]).parse();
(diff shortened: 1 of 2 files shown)


doncem:
if this improves tests, can we propagate to send letter and bulk scan processor?

apereverzin:
@rkondratowicz This check inconsistently fails. And the other check (where username and user email were being mixed) is still failing sometimes.
So this PR will not be merged.
I keep looking into this.

apereverzin:
This fix does not work

PR #384, opened 2020-03-12
Create `BlobContentExtractor`

rkondratowicz:
Extracting existing code from `BlobProcessor`.
Mostly to reduce the number of tests required after adding new method `continueProcessingEnvelope(...)` to `BlobProcessor`. (coming soon)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobContentExtractor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobContentExtractor.java
@@ -1,0 +1,39 @@
+package uk.gov.hmcts.reform.blobrouter.services;
+
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.blobrouter.config.TargetStorageAccount;
+import uk.gov.hmcts.reform.blobrouter.exceptions.InvalidZipArchiveException;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.Objects;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import static com.google.common.io.ByteStreams.toByteArray;
+import static uk.gov.hmcts.reform.blobrouter.config.TargetStorageAccount.CRIME;
+import static uk.gov.hmcts.reform.blobrouter.util.zipverification.ZipVerifiers.ENVELOPE;
+
+@Component
+public class BlobContentExtractor {
+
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobContentExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+                }
+
+                throw new InvalidZipArchiveException(
+                    String.format("ZIP file doesn't contain the required %s entry", ENVELOPE)
+                );
+            }
+        } else {
+           return blobContent;
+        }
+    }
```
doncem: maybe we could if/else in method caller instead? this is content extractor afterall

rkondratowicz: My goal was to remove the logic around whether to extract (plus how) or not to not test everything twice for both CRIME and BULKSCAN in the unit test for processor.

How about a different name for this new class then?

doncem: fair enough. we tend to struggle with names. right now can't think of one, maybe it'll kick in later :)

rkondratowicz: `TargetContentPreparator`?

doncem: :bowing_man:

nitinprabhu: `BlobContentHandler` ?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/BlobContentExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+@Component
+public class BlobContentExtractor {
+
+    public byte[] getContentToUpload(byte[] blobContent, TargetStorageAccount targetAccount) throws IOException {
+        if (targetAccount == CRIME) {
+            try (var zipStream = new ZipInputStream(new ByteArrayInputStream(blobContent))) {
+                ZipEntry entry;
+
+                while ((entry = zipStream.getNextEntry()) != null) {
+                    if (Objects.equals(entry.getName(), ENVELOPE)) {
```
lgonczar: Worth renaming `ENVELOPE` at this point, as it doesn't describe what the constant represents.

PR #386, opened 2020-03-12
rename doupdate method to update and add logs to cache methods

mdayican:
### Change description ###

rename doupdate method to update and add logs to cache methods

```
[ ] Yes
[x] No
```


lgonczar:
How about `upload` instead?

mdayican:
> How about `upload` instead?

how it became update :)

PR #390, opened 2020-03-13
Remove client for error notifications

doncem:
### Change description ###

Not used and moved to hmcts/reform-scan-notification-service#12 for centralised solution

```
[ ] Yes
[x] No
```


doncem:
configs left aside

nitinprabhu:
> configs left aside

It's worth removing all of them as single clean up PR unless there is a good reason to split them

doncem:
yes, will have to amend values.yaml and that will require chart version bump - best to leave configs aside ;). it's a reverse method. when introducing smth we add config with flux prs, then functionality. now we remove functionality and leave configs aside for the dessert

PR #393, opened 2020-03-16
Continue processing blob only if it is in 'created' status

rkondratowicz:
Extra check

Best to review in split view (wrapping existing code in `if`)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -40,6 +40,11 @@ public EnvelopeService(
         return envelopeRepository.find(blobName, containerName);
     }
 
+    @Transactional(readOnly = true)
```
lgonczar: Why is it transactional?

rkondratowicz: see all other 'read' methods in this class

lgonczar: Fine for this PR, but I wonder if that's needed.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -77,29 +80,49 @@ public void continueProcessing(UUID envelopeId, BlobClient blob) {
...
+                            handleError(id, blobClient, exception);
+                        } finally {
+                            tryToReleaseLease(leaseClient, blobClient.getBlobName(), blobClient.getContainerName());
                         }
-                    } catch (Exception exception) {
-                        handleError(id, blobClient, exception);
-                    } finally {
-                        tryToReleaseLease(leaseClient, blobClient.getBlobName(), blobClient.getContainerName());
+                    } else {
+                        logger.info(
```
lgonczar: Do we check somewhere if the envelope is marked as deleted?

rkondratowicz: no

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -77,29 +80,49 @@ public void continueProcessing(UUID envelopeId, BlobClient blob) {
             blob.getContainerName()
         );
 
-        handle(blob, () -> envelopeId);
+        handle(
+            blob,
+            () -> envelopeId,
+            new Condition(
+                () -> envelopeService.findEnvelope(envelopeId).filter(e -> e.status == Status.CREATED).isPresent(),
```
apereverzin: In case if findEnvelope returned empty result this check is misleading

rkondratowicz:
> In case if findEnvelope returned empty result the check 'status == Status.CREATED' is misleading - there is no envelope at all, not the envelope is with wrong status

@apereverzin envelope always exists

PR #395, opened 2020-03-16
Remove error notification configs

doncem:
### Change description ###

Follow up of #390 

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -7,9 +7,6 @@ java:
       resourceGroup: reform-scan
       secrets:
         - app-insights-instrumentation-key
-        - error-notifications-password
-        - error-notifications-url
-        - error-notifications-username
         - reports-email-username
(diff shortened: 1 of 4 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -7,9 +7,6 @@ java:
       resourceGroup: reform-scan
       secrets:
         - app-insights-instrumentation-key
-        - error-notifications-password
-        - error-notifications-url
-        - error-notifications-username
```
nitinprabhu: Can we also remove terraform config for copying secrets ?

doncem: i thought about it but then feared and i think my fears are right (happened for us before) - secrets will be removed. will have to quickly prep a similar pr for notification service then. if we're happy - can address that separately with both prs ready to go in and master builds are not broken/flaky

doncem:
broken pipeline

PR #404, opened 2020-03-18
Return all matching envelopes from search endpoint

rkondratowicz:
instead of just the last one


src/test/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeServiceTest.java
```diff
@@ -211,6 +213,38 @@ void should_mark_envelope_as_rejected() {
         assertThat(eventCaptor.getValue().notes).isEqualTo("some reason");
     }
 
+    @Test
+    void should_build_envelope_info() {
```
nitinprabhu: I like the way tests are written when you read it's very clear 🏅

PR #405, opened 2020-03-19
Add notification service bus queue configuration

aliveni:

### Change description ###
Flux config https://github.com/hmcts/cnp-flux-config/pull/2572

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -49,3 +49,6 @@ cftblobstorage:
   enabled: false
 crimeblobstorage:
   enabled: false
+
+servicebus:
+  enabled: false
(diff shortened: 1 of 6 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -19,6 +19,7 @@ java:
         - blob-router-POSTGRES-HOST
         - blob-router-POSTGRES-PORT
         - blob-router-POSTGRES-DATABASE
+        - notifications-queue-send-connection-string
 
   environment:
     DB_CONN_OPTIONS: ?sslmode=require
```
mdayican: `servicebus:
  enabled: false`

aliveni: updated now.

PR #407, opened 2020-03-19
Acquire lease before deciding what to do with blob

rkondratowicz:
To handle scenarios when multiple instances of blob router are 'competing' for the same file.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
@@ -2,7 +2,6 @@ package uk.gov.hmcts.reform.blobrouter.tasks.processors;
 
 import com.azure.storage.blob.BlobClient;
 import com.azure.storage.blob.models.BlobStorageException;
-import com.azure.storage.blob.specialized.BlobLeaseClient;
 import org.slf4j.Logger;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -69,17 +75,36 @@ private boolean isReady(BlobItem blob, String containerName) {
     }
 
     private void processBlob(BlobClient blob) {
-        envelopeService
-            .findLastEnvelope(blob.getBlobName(), blob.getContainerName())
+        leaseAcquirer
+            .acquireFor(blob)
```
lgonczar: It would be perfect if acquiring the lease and releasing it was done by the same class. It could be leaseAcquirer itself.

rkondratowicz: fixed

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/BlobProcessor.java
```diff
@@ -95,42 +91,29 @@ private void handle(
...
-                            blobClient.getContainerName()
-                        );
-                    }
-                },
-                () -> logger.info(
-                    "Cannot acquire a lease for blob - skipping. File name: {}, container: {}",
-                    blobClient.getBlobName(),
-                    blobClient.getContainerName()
-                )
+        if (postLeaseCondition.isMet()) {
```
lgonczar: Now that lease management is done at a higher level, `postLeaseCondition` can change name to `processingCondition` or sth. Up to you.

PR #415, opened 2020-03-20
Handle releasing lease in helper class

rkondratowicz:
Add new method to `LeaseAcquirer` that handles both acquiring and releasing a lease, so that the client does not need to worry about that.

No usage in this PR. Just new method with tests.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
@@ -23,6 +23,7 @@ public class LeaseAcquirer {
         this.leaseClientProvider = leaseClientProvider;
     }
 
+    @Deprecated
     public Optional<BlobLeaseClient> acquireFor(BlobClient blobClient) {
         try {
             var leaseClient = leaseClientProvider.get(blobClient);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
```diff
@@ -42,4 +43,41 @@ public LeaseAcquirer(LeaseClientProvider leaseClientProvider) {
             return Optional.empty();
         }
     }
+
+    public void ifAcquiredOrElse(BlobClient blobClient, Runnable onSuccess, Runnable onFailure) {
```
lgonczar: Suggestion only: how about `withLeaseOrElse`?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
```diff
@@ -42,4 +43,41 @@ public LeaseAcquirer(LeaseClientProvider leaseClientProvider) {
...
+        try {
+            var leaseClient = leaseClientProvider.get(blobClient);
+            leaseClient.acquireLease(LEASE_DURATION_IN_SECONDS);
+
+            onSuccess.run();
+
+            release(leaseClient, blobClient);
+
+        } catch (BlobStorageException exc) {
+            if (exc.getErrorCode() != BlobErrorCode.LEASE_ALREADY_PRESENT) {
```
lgonczar: If lease is already present, would it be worth logging it as info? Just wondering if that information would be useful, e.g. to see that we tried to progress with the file.

src/test/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirerTest.java
```diff
@@ -51,4 +53,42 @@ void should_return_empty_optional_if_lease_for_given_blob_has_already_been_acqui
...
+
+        // then
+        verify(onSuccess).run();
+        verify(onFailure, never()).run();
+
+        verify(leaseClient).releaseLease();
+    }
+
+    @Test
+    void should_run_provided_action_when_lease_was_not_acquired() {
```
lgonczar: Maybe `should_not_run...` or `should_run_provided_failure_action...`?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
rkondratowicz: this is copied from existing method `acquireFor`

lgonczar: Yeah, not strictly related to this PR.

PR #422, opened 2020-03-24
Added notification envelopes repository to get the rejected envelopes

aliveni:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
@@ -18,6 +18,7 @@ import java.util.UUID;
 import static java.time.Instant.now;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.catchThrowable;
+import static org.assertj.core.api.Assertions.tuple;
 
 @ActiveProfiles({"integration-test", "db-test"})
 @SpringBootTest
@@ -42,23 +43,22 @@ public class EnvelopeEventRepositoryTest {
         var event2 = new NewEnvelopeEvent(envelopeId, EventType.DELETED, "note 2");
 
         // when
-        eventRepo.insert(event1);
-        eventRepo.insert(event2);
+        long eventId1 = eventRepo.insert(event1);
+        long eventId2 = eventRepo.insert(event2);
 
         var eventsInDb = eventRepo.findForEnvelope(envelopeId);
 
         // then
-        assertThat(eventsInDb).hasSize(2);
+        assertThat(eventsInDb)
+            .hasSize(2)
+            .extracting(e -> tuple(e.id, e.envelopeId, e.type, e.notes))
+            .containsExactlyInAnyOrder(
+                tuple(eventId1, envelopeId, event1.type, event1.notes),
+                tuple(eventId2, envelopeId, event2.type, event2.notes)
+            );
 
-        assertThat(eventsInDb.get(0).id).isNotNull();
-        assertThat(eventsInDb.get(0).envelopeId).isEqualTo(event1.envelopeId);
         assertThat(eventsInDb.get(0).createdAt).isNotNull();
-        assertThat(eventsInDb.get(0).notes).isEqualTo(event1.notes);
-
-        assertThat(eventsInDb.get(1).id).isNotNull();
-        assertThat(eventsInDb.get(1).envelopeId).isEqualTo(event2.envelopeId);
         assertThat(eventsInDb.get(1).createdAt).isNotNull();
-        assertThat(eventsInDb.get(1).notes).isEqualTo(event2.notes);
     }
 
     @Test
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelope.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelope.java
@@ -1,0 +1,22 @@
+package uk.gov.hmcts.reform.blobrouter.data.notifications;
+
+public class NotificationEnvelope {
+
+    public final String container;
+    public final String fileName;
+    public final String errorType;
+    public final String notes;
+
+    public NotificationEnvelope(
+        String container,
+        String fileName,
+        String errorType,
+        String notes
+    ) {
+        this.container = container;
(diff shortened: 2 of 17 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelopeRepository.java
```diff
@@ -0,0 +1,31 @@
...
+        this.jdbcTemplate = jdbcTemplate;
+        this.mapper = mapper;
+    }
+
+    public List<NotificationEnvelope> getRejectedEnvelopes() {
+        return jdbcTemplate.query(
+            "SELECT env.file_name, env.container, event.type as error_type, event.notes as description "
+                + " FROM envelopes env, envelope_events event "
+                + " WHERE env.id = event.envelope_id "
+                + " AND env.status = 'REJECTED' AND event.type = 'REJECTED'",
```
nitinprabhu: do we need both conditions ?

aliveni: envelope status can be removed but I thought it is clear this way.

nitinprabhu: Also do we have index on status. I think we can only check on status/type and have an index ? what are your thoughts ?

aliveni: There is no index on status/type. I'll remove the status condition and add an index on `type`. is that ok?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
```diff
@@ -42,23 +43,22 @@ void should_save_and_read_events() {
...
-        eventRepo.insert(event1);
-        eventRepo.insert(event2);
+        long eventId1 = eventRepo.insert(event1);
+        long eventId2 = eventRepo.insert(event2);
 
         var eventsInDb = eventRepo.findForEnvelope(envelopeId);
 
         // then
-        assertThat(eventsInDb).hasSize(2);
+        assertThat(eventsInDb)
```
aliveni: Changed the assertion as the order of the records is different from the expectation

src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelopeRepository.java
```diff
@@ -0,0 +1,31 @@
...
+        this.jdbcTemplate = jdbcTemplate;
+        this.mapper = mapper;
+    }
+
+    public List<NotificationEnvelope> getRejectedEnvelopes() {
+        return jdbcTemplate.query(
+            "SELECT env.file_name, env.container, event.type as error_type, event.notes as description "
+                + " FROM envelopes env, envelope_events event "
+                + " WHERE env.id = event.envelope_id "
+                + " AND event.type = 'REJECTED'",
```
rkondratowicz: looks like `error_type` is always 'REJECTED' then

rkondratowicz: do we need a new model, or can this just return normal envelopes?

aliveni: Envelope doesn't have `notes`, which is mapped to `error_description` in the `NotificationMsg`

rkondratowicz: ok

rkondratowicz: `error_type` field in the model is still not needed though, right?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelope.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.hmcts.reform.blobrouter.data.notifications;
+
+public class NotificationEnvelope {
```
rkondratowicz: miho not clear name. is this  what's going to be sent to the queue?

rkondratowicz: if not, maybe `RejectedEnvelope`?

aliveni: renamed as `RejectedEnvelope`

src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelopeRepository.java
aliveni: removed `error_type`

src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelope.java
aliveni: https://github.com/hmcts/blob-router-service/blob/master/src/main/java/uk/gov/hmcts/reform/blobrouter/servicebus/notifications/model/NotificationMsg.java 
this will be sent to the queue

src/main/java/uk/gov/hmcts/reform/blobrouter/data/notifications/NotificationEnvelopeRepository.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.hmcts.reform.blobrouter.data.notifications;
+
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+
+@Repository
+public class NotificationEnvelopeRepository {
```
rkondratowicz: maybe this class should be renamed as well

aliveni: Renamed

PR #432, opened 2020-03-30
Add config for send notifications task

aliveni:

### Change description ###
Add cron expression for the task which sends messages to notifications queue.
The task should run every 10 minutes.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -31,6 +31,7 @@ java:
     CHECK_NEW_ENVELOPES_CRON: "0 0 * * * *"
     TASK_SCAN_DELAY: "4000" # in millis
     SEND_DAILY_REPORT_CRON: "0 0 18 ? * MON-FRI"
+    SEND_NOTIFICATIONS_CRON: "0 0/10 * * * *"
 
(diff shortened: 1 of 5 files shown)


aliveni:
Flux PR https://github.com/hmcts/cnp-flux-config/pull/2741

charts/reform-scan-blob-router/values.yaml
```diff
@@ -31,6 +31,7 @@ java:
     CHECK_NEW_ENVELOPES_CRON: "0 0 * * * *"
     TASK_SCAN_DELAY: "4000" # in millis
     SEND_DAILY_REPORT_CRON: "0 0 18 ? * MON-FRI"
+    SEND_NOTIFICATIONS_CRON: "0 0/10 * * * *"
```
nitinprabhu: Do we want to override for aat and preview ?

aliveni: Yeah, will update the PR

aliveni: Updated preview and aat values.yaml files

doncem: aat shouldn't be updated but since we already applied such practice for this service - let it be so

aliveni:
Closing the PR as the build is failing

PR #439, opened 2020-04-01
Add pending_notification column to envelopes table

aliveni:

### Change description ###
- Add pending_notification column to envelopes table
- Add index on pending_notification and envelope id which will be used in the rejected files query where clause.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/db/migration/V009__Add_pending_notification_column_and_index_to_envelopes.sql
+++ b/src/main/resources/db/migration/V009__Add_pending_notification_column_and_index_to_envelopes.sql
@@ -1,4 +1,4 @@
 ALTER TABLE envelopes
 ADD COLUMN pending_notification BOOLEAN DEFAULT FALSE;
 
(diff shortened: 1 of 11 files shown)


src/main/resources/db/migration/V009__Add_pending_notification_column_and_index_to_envelopes.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE envelopes
+ADD COLUMN pending_notification BOOLEAN DEFAULT FALSE;
+
+CREATE INDEX envelopes_pending_notification_envelope_id_idx ON envelopes (id, pending_notification);
```
rkondratowicz, line 4: why on those 2 columns?

aliveni: yeah, envelopes `id` is primary key so just `pending_notification` is enough. Updated the script

aliveni:
Closing this PR as the PR builds fails at AKS deployment stage

PR #440, opened 2020-04-01
Add pending_notification column to envelopes table

aliveni:
Reopening #439


aliveni:
PR build is still failing

PR #436, opened 2020-03-30
Add Notification service to send messages to the queue for the rejected envelopes

aliveni:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-1096

### Change description ###
Add Notification service
- To send messages to notifications queue for the rejected envelopes
-  Add an event when notification message is sent

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #442, opened 2020-04-01
Update envelope pending_notification and add new event NOTIFICATION_SENT

aliveni:

### Change description ###
Add methods for:
- updating `pending_notification` value in envelopes table 
- adding a new event `NOTIFICATION_SENT` for the envelope

The method added in `EnvelopeService` will be used in `NotificationService` (added in #436)

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
```diff
@@ -112,6 +112,30 @@ void should_mark_existing_envelope_as_deleted() {
...
+        Optional<Envelope> envelope = repo.find(id);
+        assertThat(envelope).hasValueSatisfying(env -> assertThat(env.pendingNotification).isEqualTo(false));
+
+        // when
+        int updateCount = repo.updatePendingNotification(id);
+        Optional<Envelope> envelopeAfterUpdate = repo.find(id);
+
+        // then
+        assertThat(updateCount).isEqualTo(1);
+        assertThat(envelopeAfterUpdate).hasValueSatisfying(env -> assertThat(env.pendingNotification).isEqualTo(true));
```
nitinprabhu, line 136: For single assertion below should be fine in my opinion but up to you
```
assertThat(envelopeAfterUpdate.get(0).pendingNotification).isEqualTo(true);

```

src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -99,6 +99,12 @@ public void markAsRejected(UUID id, String reason) {
             );
     }
 
+    @Transactional
+    public void markPendingNotificationAsSent(UUID id) {
+        envelopeRepository.updatePendingNotification(id);
+        eventRepository.insert(new NewEnvelopeEvent(id, EventType.NOTIFICATION_SENT, null));
```
nitinprabhu, line 105: it would nice to have an integration test to verify that transaction is rolled back when one fails

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
aliveni: removed the additional assertion

src/main/java/uk/gov/hmcts/reform/blobrouter/data/envelopes/EnvelopeRepository.java
```diff
@@ -136,6 +136,15 @@ public int markAsDeleted(UUID id) {
         );
     }
 
+    public int updatePendingNotification(UUID id) {
```
rkondratowicz, line 139: doesn't say what it updates it to. Maybe `markAsSent`?

rkondratowicz, line 139: doesn't say what it updates it to? Maybe `markAsPendingNotification`?

src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
rkondratowicz: if notification is sent then it's not longer pending. Pending means it should be sent but was not sent yet.

rkondratowicz: ie `is_pending` should be set to `true` when rejecting envelope and then set to `false` once it gets sent.

src/main/java/uk/gov/hmcts/reform/blobrouter/data/events/EventType.java
```diff
@@ -8,4 +8,5 @@
     DELETED_FROM_REJECTED,
     DUPLICATE_REJECTED,
     ERROR,
+    NOTIFICATION_SENT
```
rkondratowicz, line 11: trailing `,` would be nice

src/main/java/uk/gov/hmcts/reform/blobrouter/data/envelopes/EnvelopeRepository.java
aliveni: added another parameter for the value, because it is used in 2 places now (when the envelope is rejected and notification sent)

src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
aliveni: makes sense.
I updated the `markAdRejected` method to set  `pending_notification` to `true` and 
when notification is sent,  `pending_notification` will be updated to `falsea`

src/main/java/uk/gov/hmcts/reform/blobrouter/data/events/EventType.java
aliveni: added `,`

PR #443, opened 2020-04-02
Add scheduled task to send notifications for the rejected envelopes

aliveni:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/resources/application.properties
+++ b/src/integrationTest/resources/application.properties
@@ -30,6 +30,9 @@ scheduling.task.delete-rejected-files.enabled=false
 scheduling.task.reject-duplicates.enabled=false
 scheduling.task.reject-duplicates.cron=0/10 * * * * *
 
+scheduling.task.send-notifications.enabled=false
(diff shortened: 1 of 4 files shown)


aliveni:
> do we need config in main application as well? defaults will do it and next pr i presume will be the chart config alongside

the config is added in https://github.com/hmcts/blob-router-service/pull/435

src/integrationTest/resources/application.properties
```diff
@@ -30,6 +30,9 @@ scheduling.task.delete-rejected-files.enabled=false
 scheduling.task.reject-duplicates.enabled=false
 scheduling.task.reject-duplicates.cron=0/10 * * * * *
 
+scheduling.task.send-notifications.enabled=false
+scheduling.task.send-notifications.cron=0/10 * * * * *
```
nitinprabhu, line 34: is this property needed if the job is disabled ?

aliveni: Just followed the config added for existing tasks. Removed the cron property now

PR #463, opened 2020-04-21
Don't log error when blob was already deleted by another instance of a job

rkondratowicz:
The main job is not using schedlock, locks on files are used instead.
Currently, when a job tries to acquire a lease for a file that is already being process it just silently skips it.

Adding the same logic for when blob was already deleted.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
@@ -7,6 +7,8 @@ import com.azure.storage.blob.specialized.BlobLeaseClient;
 import org.slf4j.Logger;
 import org.springframework.stereotype.Component;
 
+import static com.azure.storage.blob.models.BlobErrorCode.BLOB_NOT_FOUND;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
```diff
@@ -31,7 +33,7 @@ public void ifAcquiredOrElse(BlobClient blobClient, Runnable onSuccess, Runnable
             release(leaseClient, blobClient);
 
         } catch (BlobStorageException exc) {
-            if (exc.getErrorCode() != BlobErrorCode.LEASE_ALREADY_PRESENT) {
+            if (exc.getErrorCode() != LEASE_ALREADY_PRESENT && exc.getErrorCode() != BLOB_NOT_FOUND) {
```
mdayican, line 36: in else may be we can log as warn

nitinprabhu, line 36: should the condition be OR ?

rkondratowicz: with OR it would always be true :)

rkondratowicz:
jenkins issues, closing

PR #464, opened 2020-04-21
Enable email reports in prod

apereverzin:
### Change description ###

https://github.com/hmcts/cnp-flux-config/pull/3014

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.aat.template.yaml
+++ b/charts/reform-scan-blob-router/values.aat.template.yaml
@@ -4,6 +4,7 @@ java:
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
     DELETE_DISPATCHED_FILES_CRON: "0/10 * * * * *"
     SEND_NOTIFICATIONS_CRON: "0/10 * * * * *"
+    SEND_DAILY_REPORT_ENABLED: "false"
   # Don't modify below here
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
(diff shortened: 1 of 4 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -41,8 +41,8 @@ java:
 
     BULK_SCAN_PROCESSOR_URL: http://bulk-scan-processor-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
 
-    SMTP_HOST: "false"
-    SEND_DAILY_REPORT_ENABLED: "false"
+    SEND_DAILY_REPORT_ENABLED: "true"
```
doncem, line 44: if prod only then need to go into flux config instead

apereverzin: and SMTP_HOST - also to flux?

charts/reform-scan-blob-router/values.aat.template.yaml
```diff
@@ -4,6 +4,7 @@ java:
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
     DELETE_DISPATCHED_FILES_CRON: "0/10 * * * * *"
     SEND_NOTIFICATIONS_CRON: "0/10 * * * * *"
+    SEND_DAILY_REPORT_ENABLED: "false"
```
nitinprabhu, line 7: we should not have any values here

apereverzin: yes, as Don said - changes will be only in flux config

charts/reform-scan-blob-router/values.yaml
doncem: microsoft provides that value openly so it is a nobrainer to google the value

apereverzin: I will close this PR

apereverzin:
https://github.com/hmcts/cnp-flux-config/pull/3014

PR #466, opened 2020-04-24
Make the presence of write permission in SAS token configurable

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1187

### Change description ###

Make the presence of write permission in SAS token configurable, so that it can be reduced to CREATE in a given environment.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -43,6 +43,7 @@ java:
 
     SMTP_HOST: "false"
     SEND_DAILY_REPORT_ENABLED: "false"
+    INCLUDE_WRITE_PERMISSION_IN_STORAGE_SAS_TOKEN: true
   image: 'hmctspublic.azurecr.io/reform-scan/blob-router:latest'
 reformblobstorage:
   enabled: false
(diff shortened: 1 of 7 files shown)


src/main/resources/application.yaml
```diff
@@ -85,6 +85,7 @@ flyway:
   skip-migrations: ${FLYWAY_SKIP_MIGRATIONS}
 
 storage-blob-processing-delay-in-minutes: ${STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES}
+include-write-permission-in-storage-sas-token: ${INCLUDE_WRITE_PERMISSION_IN_STORAGE_SAS_TOKEN:true}
```
nitinprabhu, line 88: this env var should be defined in `values.yaml` and that change should go first along with flux change. We can also remove the default here. Then this PR looks good to go

charts/reform-scan-blob-router/values.yaml
```diff
@@ -43,6 +43,7 @@ java:
 
     SMTP_HOST: "false"
     SEND_DAILY_REPORT_ENABLED: "false"
+    INCLUDE_WRITE_PERMISSION_IN_STORAGE_SAS_TOKEN: true
```
nitinprabhu, line 46: do we want to make this change straight on prod or we want this disable for now and enable it for only demo ?

src/main/resources/application.yaml
lgonczar: Done. Flux change was always planned for this bit, now in https://github.com/hmcts/cnp-flux-config/pull/3088.

charts/reform-scan-blob-router/values.yaml
lgonczar: `true` is what currently happens.

PR #487, opened 2020-05-05
Add indexes

rkondratowicz:
For:
- checking if envelope exists (first step in processing every blob)
- finding envelopes for which blob can be deleted (scheduled job that cleans up the blob storage based on information found in DB)

https://tools.hmcts.net/jira/browse/BPS-1235

### DIFF
--- a/src/main/resources/db/migration/V011__Add_index_for_checking_if_envelope_exists.sql
+++ b/src/main/resources/db/migration/V011__Add_index_for_checking_if_envelope_exists.sql
@@ -1,0 +1,2 @@
+CREATE INDEX envelopes_file_name_container_idx
+ON envelopes (file_name, container);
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/V011__Add_index_for_checking_if_envelope_exists.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX envelopes_file_name_container_idx
+ON envelopes (file_name, container);
```
lgonczar, line 2: How about reverting the order of the two columns in the index? This way we could use the index when searching for envelopes for the given container.

rkondratowicz: good point

PR #488, opened 2020-05-05
Added staging db infra config for functional tests to be used in AAT

nitinprabhu:

https://tools.hmcts.net/jira/browse/BPS-1216

### Change description ###

- Creating database to be used for functional tests by staging pod.

- Although this would also create DB in PROD environment it won't be used as we don't have functional tests or staging pod in prod.

- Next PR would make use of this 

```
[ ] Yes
[ x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -21,6 +21,21 @@ module "reform-blob-router-db" {
   subscription       = "${var.subscription}"
 }
 
+# Staging DB to be used by AAT staging pod for functional tests
+module "reform-blob-router-staging-db" {
+  source             = "git@github.com:hmcts/cnp-module-postgres?ref=master"
+  product            = "${var.component}-staging"
+  location           = "${var.location_db}"
+  env                = "${var.env}"
+  database_name      = "blob_router"
+  postgresql_user    = "blob_router"
+  postgresql_version = "11"
+  sku_name           = "GP_Gen5_2"
+  sku_tier           = "GeneralPurpose"
+  common_tags        = "${var.common_tags}"
+  subscription       = "${var.subscription}"
+}
+
 # region: key vault definitions
 
 data "azurerm_key_vault" "bulk_scan_key_vault" {
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -142,6 +142,38 @@ resource "azurerm_key_vault_secret" "POSTGRES_DATABASE" {
 }
 # endregion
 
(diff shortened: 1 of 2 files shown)


infrastructure/main.tf
```diff
@@ -127,6 +142,38 @@ resource "azurerm_key_vault_secret" "POSTGRES_DATABASE" {
...
+}
+
+resource "azurerm_key_vault_secret" "STAGING_POSTGRES_PORT" {
+  name         = "${var.component}-staging-POSTGRES-PORT"
+  key_vault_id = "${data.azurerm_key_vault.reform_scan_key_vault.id}"
+  value        = "${module.reform-blob-router-staging-db.postgresql_listen_port}"
+}
+
+resource "azurerm_key_vault_secret" "STAGING_POSTGRES_DATABASE" {
+  name         = "${var.component}--stagingPOSTGRES-DATABASE"
```
mdayican, line 171: typo

mdayican: and we do not need used capitals for staging

doncem, line 171: typo - missing `-`

doncem, line 171: will those `resource` values and names be used by jenkins pipeline or any other automatic process? would be really nice to sort those names out for once:

- correct usage of case
- correct usage of `_` vs `-`
- no need to reference type of db e.g. `POSTGRES`

i know it is hardcoded for main pipelines we have but if we are free to fix it here - i suggest to do it

nitinprabhu: > could use `count` for both `module` and `resource` declarations

Already discussed why we can't during refinement :)

nitinprabhu: keeping those consistent with what we have

doncem: ok. certain sentences can slip through as kid is climbing on my head all the time during the meetings. that's his participation option :D

doncem: i know about consistency ;). just saying politely suggesting it is not correct. which is not your fault anyway

nitinprabhu: ok kept those following what we have but can update as @doncem thinks the same

nitinprabhu: problem with introducing count in upstream module is it might have breaking changes. That's was our first choice but platform too recommended doing it this way :)

nitinprabhu: sure will do it :)

PR #491, opened 2020-05-06
Search for envelope in DB after the lease has been acquired

lgonczar:
### Change description ###

Tests still to be added.

Search for envelope in DB after the lease has been acquired on blob. Even when we're doing it before acquiring the lease, we still need to search after. Otherwise, we'd allow for race conditions.

Both checks are necessary - the first one is to avoid acquiring lease on the blob and the second one to get the accurate state.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
@@ -84,18 +84,20 @@ public class ContainerProcessor {
                 envelope -> logger.info(
                     "Envelope already processed in system, skipping. {} ", envelope.getBasicInfo()
                 ),
-                () -> handleBlob(optionalEnvelope, blob)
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -84,18 +84,20 @@ private void processBlob(BlobClient blob) {
                 envelope -> logger.info(
                     "Envelope already processed in system, skipping. {} ", envelope.getBasicInfo()
                 ),
-                () -> handleBlob(optionalEnvelope, blob)
+                () -> handleBlob(blob)
             );
     }
 
-    private void handleBlob(Optional<Envelope> optionalEnvelope, BlobClient blob) {
+    private void handleBlob(BlobClient blob) {
```
rkondratowicz, line 91: this has already been change, this branch needs refreshing

PR #495, opened 2020-05-07
Update envelope service method to return envelope events

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1222

### Change description ###
Envelope service method should return envelope events along with envelope details.
In the next PR, Envelopes controller will use this method.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
@@ -17,6 +17,7 @@ import uk.gov.hmcts.reform.blobrouter.data.events.NewEnvelopeEvent;
 import java.util.UUID;
 
 import static java.time.Instant.now;
+import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.catchThrowable;
 import static org.assertj.core.api.Assertions.tuple;
@@ -80,4 +81,46 @@ public class EnvelopeEventRepositoryTest {
     void should_return_empty_list_when_there_are_no_events_for_given_envelope() {
         assertThat(eventRepo.findForEnvelope(UUID.randomUUID())).isEmpty();
     }
+
+    @Test
+    void should_return_events_for_envelope_ids_order_by_created_at() {
+        // given
+        var envelopeId1 = envelopeRepo.insert(new NewEnvelope("c1", "f1", now(), null, Status.CREATED));
+        var envelopeId2 = envelopeRepo.insert(new NewEnvelope("c2", "f2", now(), null, Status.DISPATCHED));
+
+        var event1a = new NewEnvelopeEvent(envelopeId1, EventType.FILE_PROCESSING_STARTED, null, "note 1");
+        var event2a = new NewEnvelopeEvent(envelopeId2, EventType.FILE_PROCESSING_STARTED, null, "note 3");
+        var event2b = new NewEnvelopeEvent(envelopeId2, EventType.DISPATCHED, null, "note 4");
+
+        // when
+        var eventId1a = eventRepo.insert(event1a);
+        var eventId2a = eventRepo.insert(event2a);
+        var eventId2b = eventRepo.insert(event2b);
+
+        var eventsInDb = eventRepo.findForEnvelopes(asList(envelopeId1, envelopeId2));
+
+        // then
+        assertThat(eventsInDb)
+            .hasSize(3)
+            .extracting(e -> tuple(e.id, e.envelopeId, e.type, e.errorCode, e.notes))
+            .containsExactly(
+                tuple(eventId2b, envelopeId2, event2b.type, event2b.errorCode, event2b.notes),
+                tuple(eventId2a, envelopeId2, event2a.type, event2a.errorCode, event2a.notes),
+                tuple(eventId1a, envelopeId1, event1a.type, event1a.errorCode, event1a.notes)
+            );
+    }
+
+    @Test
+    void should_return_empty_when_no_events_exist_for_envelopes() {
+        // given
+        var envelopeId1 = envelopeRepo.insert(new NewEnvelope("c1", "f1", now(), null, Status.CREATED));
+        var envelopeId2 = envelopeRepo.insert(new NewEnvelope("c2", "f2", now(), null, Status.CREATED));
+
+        // when
+        var eventsInDb = eventRepo.findForEnvelopes(asList(envelopeId1, envelopeId2));
+
+        // then
+        assertThat(eventsInDb).isEmpty();
+    }
+
 }
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.blobrouter.data;
 
+import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -55,8 +56,8 @@ public class EnvelopeEventRepositoryTest {
             .hasSize(2)
             .usingElementComparatorIgnoringFields("createdAt")
             .containsExactlyInAnyOrder(
-                new EnvelopeEvent(eventId1, envelopeId, event1.type, event1.errorCode, event1.notes, now()),
-                new EnvelopeEvent(eventId2, envelopeId, event2.type, event2.errorCode, event2.notes, now())
+                envelopeEvent(envelopeId, event1, eventId1),
+                envelopeEvent(envelopeId, event2, eventId2)
             );
 
         assertThat(eventsInDb.get(0).createdAt).isNotNull();
@@ -104,9 +105,9 @@ public class EnvelopeEventRepositoryTest {
             .hasSize(3)
             .usingElementComparatorIgnoringFields("createdAt")
             .containsExactly(
-                new EnvelopeEvent(eventId2b, envelopeId2, event2b.type, event2b.errorCode, event2b.notes, now()),
-                new EnvelopeEvent(eventId2a, envelopeId2, event2a.type, event2a.errorCode, event2a.notes, now()),
-                new EnvelopeEvent(eventId1a, envelopeId1, event1a.type, event1a.errorCode, event1a.notes, now())
+                envelopeEvent(envelopeId2, event2b, eventId2b),
+                envelopeEvent(envelopeId2, event2a, eventId2a),
+                envelopeEvent(envelopeId1, event1a, eventId1a)
             );
     }
 
@@ -121,6 +122,11 @@ public class EnvelopeEventRepositoryTest {
 
         // then
         assertThat(eventsInDb).isEmpty();
+    }
+
+    @NotNull
+    private EnvelopeEvent envelopeEvent(UUID envelopeId, NewEnvelopeEvent event, long eventId) {
+        return new EnvelopeEvent(eventId, envelopeId, event.type, event.errorCode, event.notes, now());
     }
 
 }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -132,7 +132,18 @@ public void saveEvent(UUID envelopeId, EventType eventType) {
     }
 
     @Transactional(readOnly = true)
-    public List<Envelope> getEnvelopes(String blobName, String containerName, LocalDate date) {
-        return envelopeRepository.findEnvelopes(blobName, containerName, date);
+    public List<Tuple2<Envelope, List<EnvelopeEvent>>> getEnvelopes(
```
lgonczar, line 135: Can we perhaps create a special class to represent envelope info?

aliveni: We already have `EnvelopeInfo` class. the result will be mapped to `EnvelopeInfo` class in the controller

lgonczar: fair enough

lgonczar:
It would be good to apply order to envelopes and events in those envelopes. I propose ordering envelopes by creation time desc, and events by id desc.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
aliveni: Do you think it would be better to replace this by a "join" on envelopes and envelope_events and have a mapper class?

lgonczar: It would be one query instead of hundreds, so a good idea.

lgonczar: Either a join with a mapper or two queries - one retrieving envelopes and the other retrieving all events for those envelopes.

aliveni: cool, I'll update the pr

aliveni: I've updated to get the data in two queries.

src/main/java/uk/gov/hmcts/reform/blobrouter/data/events/EnvelopeEventRepository.java
```diff
@@ -28,6 +28,14 @@ public EnvelopeEventRepository(NamedParameterJdbcTemplate jdbcTemplate, Envelope
         );
     }
 
+    public List<EnvelopeEvent> findForEnvelopes(List<UUID> envelopeIds) {
+        return jdbcTemplate.query(
+            "SELECT * FROM envelope_events WHERE envelope_id in (:envelopeIds) ORDER BY id DESC",
```
doncem, line 33: probably should make sure its not empty list. maybe even use specific class for that

aliveni: Good point! I'll update

lgonczar: Is it needed? In blob-router, unlike in bulk-scan-processor, envelopes always exist.

aliveni: but there may be not any envelopes matching for the search criteria (file_name, container and date)

lgonczar: Aaa, input list. Good point. Probably best to do the check in the service.

aliveni: Updated service method to check if envelopes list is empty

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
```diff
@@ -80,4 +81,46 @@ void should_throw_exception_when_trying_to_create_an_event_for_not_existing_enve
...
+
+    @Test
+    void should_return_events_for_envelope_ids_order_by_created_at() {
+        // given
+        var envelopeId1 = envelopeRepo.insert(new NewEnvelope("c1", "f1", now(), null, Status.CREATED));
+        var envelopeId2 = envelopeRepo.insert(new NewEnvelope("c2", "f2", now(), null, Status.DISPATCHED));
+
+        var event1a = new NewEnvelopeEvent(envelopeId1, EventType.FILE_PROCESSING_STARTED, null, "note 1");
+        var event2a = new NewEnvelopeEvent(envelopeId2, EventType.FILE_PROCESSING_STARTED, null, "note 3");
+        var event2b = new NewEnvelopeEvent(envelopeId2, EventType.DISPATCHED, null, "note 4");
```
lgonczar, line 93: I wonder if we need to make a minimal pause (like a millisecond) to guarantee different creation data

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
```diff
@@ -80,4 +81,46 @@ void should_throw_exception_when_trying_to_create_an_event_for_not_existing_enve
...
+
+        // then
+        assertThat(eventsInDb)
+            .hasSize(3)
+            .extracting(e -> tuple(e.id, e.envelopeId, e.type, e.errorCode, e.notes))
+            .containsExactly(
+                tuple(eventId2b, envelopeId2, event2b.type, event2b.errorCode, event2b.notes),
+                tuple(eventId2a, envelopeId2, event2a.type, event2a.errorCode, event2a.notes),
+                tuple(eventId1a, envelopeId1, event1a.type, event1a.errorCode, event1a.notes)
+            );
```
lgonczar, line 110: Would it be possible to use `usingFieldByFieldComparator` instead of tuples?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
```diff
@@ -331,15 +331,16 @@ void should_return_envelopes_count_for_container_and_time_range() {
     @Test
     void should_return_envelopes_for_the_requested_date() {
         //given
-        addEnvelope("C1", "f1");
-        addEnvelope("C3", "f2");
-        addEnvelope("C3", "f3");
+        UUID id1 = addEnvelope("f1", "C1");
+        UUID id2 = addEnvelope("f2", "C2");
+        UUID id3 = addEnvelope("f3", "C2");
```
lgonczar, line 336: Not sure how stable this test will be if we save envelopes one after another. Can they be saved within the same millisecond?

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
aliveni: Event id can not be asserted as it is auto-generated

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeRepositoryTest.java
aliveni: I'll add `Thread.sleep()` with 2 milliseconds

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
lgonczar: How about something like `usingElementComparatorIgnoringFields("id")`? Tuples are not super readable - that's where I'm coming from.

aliveni: sure, I wasn't aware that this method exists. I'll update the tests.
Also, I just noticed that it is the `createdAt` field that can't be asserted as it is saved with `CURRENT_TIMESTAMP`.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -132,7 +135,32 @@ public void saveEvent(UUID envelopeId, EventType eventType) {
...
+
+            Map<UUID, List<EnvelopeEvent>> eventsByEnvelopeIds = envelopeEvents
+                .stream()
+                .collect(groupingBy(envelopeEvent -> envelopeEvent.envelopeId));
+
+            return envelopes
+                .stream()
+                .map(envelope -> Tuples.of(
+                    envelope,
+                    eventsByEnvelopeIds.get(envelope.id)
```
lgonczar, line 161: If, for some reason, there are no events for the given envelope, this will be null. Should we make it an empty list in such case?

aliveni: Updated the service to return the empty list when no events exist for envelope

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
```diff
@@ -52,10 +53,10 @@ void should_save_and_read_events() {
         // then
         assertThat(eventsInDb)
             .hasSize(2)
-            .extracting(e -> tuple(e.id, e.envelopeId, e.type, e.errorCode, e.notes))
+            .usingElementComparatorIgnoringFields("createdAt")
             .containsExactlyInAnyOrder(
-                tuple(eventId1, envelopeId, event1.type, event1.errorCode, event1.notes),
```
lgonczar, line 57: This probably shouldn't be any order anymore. We might also need to add a pause between event insertions.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
```diff
@@ -80,4 +81,46 @@ void should_throw_exception_when_trying_to_create_an_event_for_not_existing_enve
...
+        var envelopeId2 = envelopeRepo.insert(new NewEnvelope("c2", "f2", now(), null, Status.DISPATCHED));
+
+        var event1a = new NewEnvelopeEvent(envelopeId1, EventType.FILE_PROCESSING_STARTED, null, "note 1");
+        var event2a = new NewEnvelopeEvent(envelopeId2, EventType.FILE_PROCESSING_STARTED, null, "note 3");
+        var event2b = new NewEnvelopeEvent(envelopeId2, EventType.DISPATCHED, null, "note 4");
+
+        // when
+        var eventId1a = eventRepo.insert(event1a);
+        var eventId2a = eventRepo.insert(event2a);
+        var eventId2b = eventRepo.insert(event2b);
```
lgonczar, line 98: We probably need to add pauses between envelope and event insertions.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
aliveni: This is an existing test which is using `tuple`. I updated it to use `usingElementComparatorIgnoringFields`. I'll revert these changes if you want them in a separate PR

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
```diff
@@ -80,4 +81,46 @@ void should_throw_exception_when_trying_to_create_an_event_for_not_existing_enve
...
+
+        var eventsInDb = eventRepo.findForEnvelopes(asList(envelopeId1, envelopeId2));
+
+        // then
+        assertThat(eventsInDb)
+            .hasSize(3)
+            .usingElementComparatorIgnoringFields("createdAt")
+            .containsExactly(
+                new EnvelopeEvent(eventId2b, envelopeId2, event2b.type, event2b.errorCode, event2b.notes, now()),
```
lgonczar, line 107: Suggestion, just to consider: we could extract a method, called `envelopeEvent(NewEnvelopeEvent newEnvelopeEvent)` and use it here, referencing passing `event1a`, `event2a` and `event2b` as the only argument. This could be a bit clearer, but it's up to you.

lgonczar:
Please disregard all my comments regarding pauses - timestamp has microsecond precision.

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/data/EnvelopeEventRepositoryTest.java
aliveni: Refactored but it needed more params envelope id and event id (auto-generated)

PR #504, opened 2020-05-13
Added secret mapping and bootstrap config for staging infrastructure

nitinprabhu:

https://tools.hmcts.net/jira/browse/BPS-1216

### Change description ###

- This PR overrides database , blob storage and queue connection string secrets with staging infra ones. 

- Other secrets remain same but needs to be present while overriding.

- This PR is dependent on https://github.com/hmcts/reform-scan-shared-infra/pull/43 to create secrets in key vault.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -30,7 +30,7 @@ def combinedSecrets = [
      secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
 
      secret('storage-account-secondary-key', 'TEST_STORAGE_ACCOUNT_KEY'),
-     secret('storage-account-name', 'TEST_STORAGE_ACCOUNT_NAME'),
+     secret('storage-account-staging-secondary-key', 'TEST_STORAGE_ACCOUNT_NAME'),
    ]
  ]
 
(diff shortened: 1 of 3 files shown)


nitinprabhu:
> how will the migrations work? i assume both staging and non staging dbs will have to be updated

good point missed it. @mdayican answered it :)

mdayican:
I guess functional test will fail in aat staging as they are still using aat infra but app will run in staging .
Jenkinsfile_CNP files need to be configured.

nitinprabhu:
> I guess functional test will fail in aat staging as they are still using aat infra but app will run in staging .
> Jenkinsfile_CNP files need to be configured.

good catch @mdayican

Jenkinsfile_CNP
```diff
@@ -88,7 +88,7 @@ withPipeline(type, product, component) {
 
   // Vars needed for smoke / functional testing
   env.TEST_STORAGE_CONTAINER_NAME = 'bulkscan'
-  env.TEST_STORAGE_ACCOUNT_URL = 'https://reformscanaat.blob.core.windows.net'
+  env.TEST_STORAGE_ACCOUNT_URL = 'https://reformscanaatstaging.blob.core.windows.net'
```
mdayican, line 91: may we need 2 different values for pr and nonpr ?

nitinprabhu: This is used for non PR functional tests. For PR's we use OSBA

PR #507, opened 2020-05-14
Save exception message in event notes

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeService.java
```diff
@@ -117,6 +117,11 @@ public void markEnvelopeAsDeleted(Envelope envelope) {
         eventRepository.insert(new NewEnvelopeEvent(envelope.id, EventType.DELETED, null, null));
     }
 
+    @Transactional
+    public void saveEvent(UUID envelopeId, EventType eventType, String notes) {
+        eventRepository.insert(new NewEnvelopeEvent(envelopeId, eventType, null, notes));
```
rkondratowicz, line 122: notes column in of type `TEXT`, so no need to trim

PR #515, opened 2020-05-18
Added logging to blob-dispatcher job

apereverzin:

### Change description ###
Added logging to get more information in case of similar exception in the future.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProxy.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProxy.java
@@ -61,7 +61,7 @@ public class BlobContainerClientProxy {
                     .getBlobClient(blobName)
                     .getBlockBlobClient();
 
-            logger.info("Uploading blob content to Container: {}", destinationContainer);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProxy.java
```diff
@@ -50,13 +56,20 @@ public void upload(
...
         try {
-            get(targetStorageAccount, destinationContainer)
-                .getBlobClient(blobName)
-                .getBlockBlobClient()
+            final BlockBlobClient blockBlobClient =
+                get(targetStorageAccount, destinationContainer)
+                    .getBlobClient(blobName)
+                    .getBlockBlobClient();
+
+            logger.info("Uploading blob content to Container: {}", destinationContainer);
```
lgonczar, line 64: Worth adding blob name to this log entry, as well.

lgonczar: And to the 'finished' entry.

apereverzin: Have done

PR #518, opened 2020-05-20
[DO NOT MERGE] Standalone app to run blob dispatcher

apereverzin:

### Change description ###
This PR is for investigation purposes only, not to be merged or deployed. It will fail in the pipeline because a lot of dependencies have been disabled.

Its intent is to run scheduled BlobDispatcherTask frequently (with delay 100ms) and check if throwing a RuntimeException while 'processing' each container will cause failure of the task.

To run the project as a standalone application start postgres server locally, create database called db and type 

`./gradlew bootRun`

in command line.

```
[ ] Yes
[x] No
```


apereverzin:
Used for tests

PR #519, opened 2020-05-20
Added ilbIp var as the terraform module enforces to set the value

nitinprabhu:


nitinprabhu:
Not needed anymore as the pipeline has removed it

PR #521, opened 2020-05-21
[DO NOT MERGE] Test exception

apereverzin:

### Change description ###
Trying to simulate job failure due to exception

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
@@ -33,7 +33,8 @@ public class BlobDispatcherTask {
         this.containers = serviceConfiguration.getEnabledSourceContainers();
     }
 
-    @Scheduled(fixedDelayString = "${scheduling.task.scan.delay}")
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/BlobDispatcherTask.java
```diff
@@ -33,7 +33,8 @@ public BlobDispatcherTask(
         this.containers = serviceConfiguration.getEnabledSourceContainers();
     }
 
-    @Scheduled(fixedDelayString = "${scheduling.task.scan.delay}")
+    //@Scheduled(fixedDelayString = "${scheduling.task.scan.delay}")
+    @Scheduled(fixedDelayString = "50")
```
lgonczar, line 37: How about throwing the exception right after logging the start of the job, so that nothing else happens in this job?

apereverzin: @lgonczar In this case we have multiple exceptions per one job run. Let's have a look at how existing code behaves, after some time I will add throwing exception immediately.

apereverzin:
Used for tests

PR #524, opened 2020-05-22
Delete blob along with snapshots

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1252


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/RejectedContainerCleaner.java
```diff
@@ -73,7 +75,10 @@ private void delete(BlobClient blobClient) {
         );
 
         try {
-            blobClient.delete();
+            // every time a duplicate is moved to rejected container
+            // a snapshot is created and original blob is replaced,
+            // therefore snapshots are always older than the 'base' blob and it is safe to delete them
+            blobClient.deleteWithResponse(DeleteSnapshotsOptionType.INCLUDE, null, null, Context.NONE);
```
rkondratowicz, line 81: ` blobClient.delete()` did this:
```java
    public void delete() {
        deleteWithResponse(null, null, null, Context.NONE);
    }
```

PR #526, opened 2020-05-26
Always include 'create' permissions instead of 'write' in SAS tokens

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1187

### Change description ###

Always include `create` permissions instead of `write` in SAS tokens. `Create` permission has proven to be sufficient in prod, so no need to configure it.

Next: cleanup in flux config.

```
[ ] Yes
[x] No
```


lgonczar:
> pipeline certainly already complained about chart bump, just noting here ;)

What do you mean?

doncem:
> > pipeline certainly already complained about chart bump, just noting here ;)
> What do you mean?

I mean pipeline definitely reported (didn't check) the miss on chart version bump because `values.yml` has changed. So. Just noted. Certainly you've seen it

PR #530, opened 2020-05-29
Set proxy for blob storage client as it will be rejected otherwise

nitinprabhu:
### Change description ###

- Set proxy for functional tests so that all requests goes through proxy and they don't get blocked due NSG rules.

```
[ ] Yes
[ x] No
```


nitinprabhu:
Closing this as there is 
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: reform-scan, name: servicebus-namespace-reform-scan-blob-router-pr-530
```

PR #531, opened 2020-05-29
Set proxy for blob storage client as it will be rejected otherwise

nitinprabhu:
### Change description ###

- Set proxy for functional tests so that all requests goes through proxy and they don't get blocked due NSG rules.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/FunctionalTestBase.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/FunctionalTestBase.java
@@ -1,5 +1,7 @@
 package uk.gov.hmcts.reform.blobrouter;
 
+import com.azure.core.http.ProxyOptions;
+import com.azure.core.http.netty.NettyAsyncHttpClientBuilder;
 import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.BlobServiceClientBuilder;
 import com.azure.storage.common.StorageSharedKeyCredential;
@@ -7,6 +9,7 @@ import io.restassured.RestAssured;
 import uk.gov.hmcts.reform.blobrouter.config.TestConfiguration;
 import uk.gov.hmcts.reform.blobrouter.data.envelopes.Status;
 
+import java.net.InetSocketAddress;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.concurrent.ThreadLocalRandom;
@@ -32,6 +35,19 @@ public abstract class FunctionalTestBase {
         );
 
         this.blobRouterStorageClient = new BlobServiceClientBuilder()
+            .httpClient(
+                new NettyAsyncHttpClientBuilder()
+                    .proxy(
+                        new ProxyOptions(
+                            ProxyOptions.Type.HTTP,
+                            new InetSocketAddress(
+                                "proxyout.reform.hmcts.net",
+                                8080
+                            )
+                        )
+                    )
+                    .build()
+            )
             .credential(blobRouterStorageCredential)
             .endpoint(config.sourceStorageAccountUrl)
             .buildClient();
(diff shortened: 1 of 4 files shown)


src/functionalTest/resources/application.conf
```diff
@@ -3,5 +3,6 @@ blob-router-url=${TEST_URL}
 source-storage-account-name=${TEST_STORAGE_ACCOUNT_NAME}
 source-storage-account-key=${TEST_STORAGE_ACCOUNT_KEY}
 source-storage-account-url=${TEST_STORAGE_ACCOUNT_URL}
+use-proxy-for-source-storage=${TEST_USE_PROXY_FOR_SOURCE_STORAGE}
```
nitinprabhu, line 6: we need to set this in Jenkins_cnp to false

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/FunctionalTestBase.java
```diff
@@ -31,7 +34,25 @@ protected void setUp() {
             config.sourceStorageAccountKey
         );
 
-        this.blobRouterStorageClient = new BlobServiceClientBuilder()
+        BlobServiceClientBuilder blobServiceClientBuilder = new BlobServiceClientBuilder();
+
+        if (Boolean.TRUE.equals(config.useProxyForSourceStorage)) {
```
nitinprabhu, line 39: this won't work ? 

```Boolean.valueOf(config.storageProxyEnabled) == true```

Comment (doncem) [reply src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/FunctionalTestBase.java:39]: Probably. Just drop the `== true` bit if going to use

Comment (rkondratowicz) [reply src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/FunctionalTestBase.java:39]: why not just `if(!config.useProxyForSourceStorage)`?

File: src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/config/TestConfiguration.java [hunk 3]
```diff
@@ -11,6 +11,7 @@
     public final String sourceStorageAccountKey;
     public final String sourceStorageAccountUrl;
     public final String crimeSourceContainer;
+    public final Boolean useProxyForSourceStorage;
```
rkondratowicz, line 14: why `Boolean` and not `boolean`?

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/FunctionalTestBase.java
mdayican: I thought it can be null. But seems conf does not support null value.

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/config/TestConfiguration.java
mdayican: I want it to be null as well.

rkondratowicz: What's the reason for that?

Also in constructor `boolean getBoolean(String path);` is called to set this field, so I think it is never null.

mdayican: to avoid setting TEST_USE_PROXY_FOR_SOURCE_STORAGE in cnp jenkins. as I said in my comment conf does not support null.

src/functionalTest/java/uk/gov/hmcts/reform/blobrouter/FunctionalTestBase.java
```diff
@@ -31,7 +34,27 @@ protected void setUp() {
...
+                            new InetSocketAddress(
+                                "proxyout.reform.hmcts.net",
+                                8080
+                            )
+                        )
+                    )
+                    .build()
+            );
+
+            System.out.println("PROXY set for blobRouterStorageClient");
```
rkondratowicz, line 54: can this be removed?

mdayican: I will but let me test it and see it in post func test.

apereverzin: @mdayican As far as I understand this is supposed to be removed

PR #542, opened 2020-06-01
fix dependabot dependencies and suppression

mdayican:
### Change description ###

fix dependabot dependencies and suppression

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ plugins {
   id 'org.owasp.dependencycheck' version '5.3.2.1'
   id 'com.github.ben-manes.versions' version '0.28.0'
   id 'org.sonarqube' version '3.0'
-  id 'org.flywaydb.flyway' version '6.4.2'
+  id 'org.flywaydb.flyway' version '6.4.3'
 }
 
 group = 'uk.gov.hmcts.reform'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -243,18 +243,18 @@ dependencies {
     exclude group: 'junit', module: 'junit'
     exclude group: 'org.junit.vintage', module: 'junit-vintage-engine'
   }
-  testImplementation group: 'com.azure', name: 'azure-core-test', version: '1.1.0', {
+  testImplementation group: 'com.azure', name: 'azure-core-test', version: '1.2.1', {
```
doncem, line 246: i don't think we are using it at all ;)

build.gradle
```diff
@@ -243,18 +243,18 @@ dependencies {
...
     exclude group: 'com.github.tomakehurst', module: 'wiremock-standalone'
   }
 
   testCompile group: 'org.apache.commons', name: 'commons-email', version: '1.5'
 
   integrationTestImplementation sourceSets.main.runtimeClasspath
   integrationTestImplementation sourceSets.test.runtimeClasspath
-  integrationTestImplementation group: 'com.azure', name: 'azure-core-test', version: '1.1.0', {
+  integrationTestImplementation group: 'com.azure', name: 'azure-core-test', version: '1.2.1', {
```
doncem, line 254: weird. shouldn't be here if it is in `testImplementation` scope

build.gradle
mdayican: we are using. but as I see we can get rid of. we used in mock.

doncem: so yea - we are not using it :D

build.gradle
mdayican: I thought it is on purpose. removed it.

build.gradle
mdayican: we are using a mock object from the lib.

PR #558, opened 2020-06-15
Do not release lease for deleted blobs

doncem:

[Acquire lease before deleting blob](https://tools.hmcts.net/jira/browse/BPS-1205)

### Change description ###

Deleting blob with lease it destroys it on the fly. Basically we don't need to make extra call to release the lease as blob is already deleted. Trace from app insights:

| Key | Value |
| ----- | -------- |
| BlobStorageException | Status code 404, "&lt;?xml version="1.0" encoding="utf-8"?&gt;&lt;Error&gt;&lt;Code&gt;BlobNotFound&lt;/Code&gt;&lt;Message&gt;The specified blob does not exist.<br>RequestId:2f53686c-601e-0004-71fb-42563d000000<br>Time:2020-06-15T10:00:00.5210724Z&lt;/Message&gt;&lt;/Error&gt;" |
| Logger Message | Could not release the lease with ID 486de9d0-c07b-451d-8223-d0f2d1161ca7. Blob: 2016704010065_15-06-2020-09-48-18.zip, container: probate |
| LoggerName | uk.gov.hmcts.reform.blobrouter.services.storage.LeaseAcquirer |
| LoggingLevel | WARN |

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
@@ -26,29 +26,17 @@ public class LeaseAcquirer {
     }
 
     /**
-     * Main wrapper for blobs to be leased by {@link BlobLeaseClient} and perform non-delete task.
-     * @param blobClient Represents blob
-     * @param onSuccess Runnable task to perform when lease is acquired
-     * @param onFailure Extra step to execute in case an error occurred
-     */
-    public void processAndRelease(BlobClient blobClient, Runnable onSuccess, Consumer<BlobErrorCode> onFailure) {
-        var leaseClient = ifAcquiredOrElse(blobClient, leaseId -> onSuccess.run(), onFailure);
-
-        if (leaseClient != null) {
-            release(leaseClient, blobClient);
-        }
-    }
-
-    /**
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
```diff
@@ -26,12 +26,26 @@ public LeaseAcquirer(LeaseClientProvider leaseClientProvider) {
...
+    }
+
+    /**
+     * Main wrapper for blobs to be leased by {@link BlobLeaseClient} and perform deletion task.
      * @param blobClient Represents blob
      * @param onSuccess Consumer which takes in {@code leaseId} acquired with {@link BlobLeaseClient}
      * @param onFailure Extra step to execute in case an error occurred
      */
-    public void ifAcquiredOrElse(
+    public BlobLeaseClient ifAcquiredOrElse(
```
rkondratowicz, line 48: How about simply adding a boolean parameter to this method?

Then no `processAndRelease` method (that does not even actually always release) will be needed and this class will not be concerned what kind of task is being done (new javadoc "and perform non-delete task" / "and perform non-delete task."

doncem: ok

PR #561, opened 2020-06-18
Fix envelopes datetime fields format with Local timezone

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1260

### Change description ###
Date time fields in the `/envelopes` endpoint still return in UTC format. 
Use `DatetimeFormatter` to format the values properly.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeController.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeController.java
@@ -73,7 +73,7 @@ public class EnvelopeController {
 
     private String toLocalTimeZone(Instant instant) {
         if (instant != null) {
-            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeController.java
```diff
@@ -69,9 +71,10 @@ private EnvelopeInfo toResponse(Envelope dbEnvelope, List<EnvelopeEvent> dbEvent
         );
     }
 
-    private Instant toLocalTimeZone(Instant instant) {
+    private String toLocalTimeZone(Instant instant) {
         if (instant != null) {
-            return instant.atZone(EUROPE_LONDON_ZONE_ID).toInstant();
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
```
rkondratowicz, line 76: previously we had 'T' between date and time, to follow ISO standard https://en.wikipedia.org/wiki/ISO_8601

aliveni: Updated the format

PR #575, opened 2020-06-29
Escape possible html message

doncem:
### Change description ###

E2E tests fail to walk the tree when html is present in there. Quick solution and seems like silent agreement - just to escape it.

```
[ ] Yes
[x] No
```


doncem:
merge happened to fast - messed up resources for pr build - need a fresh start. sorry :(

PR #581, opened 2020-07-14
add count to search result and add test

mdayican:
### Change description ###

- > add count to search result 
- > add test for empty result
- > use existing util method for formatting

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/SearchResult.java
```diff
@@ -6,10 +6,14 @@
...
 
+    @JsonProperty("count")
+    public final int count;
+
     @JsonProperty("data")
     public final List<?> data;
 
     public SearchResult(List<?> data) {
         this.data = data;
+        count = (data == null) ? 0 : data.size();
```
nitinprabhu, line 17: can use `CollectionUtils.size` which will handle even null :) but up to you

mdayican: which jar is it? I think it is not there.

nitinprabhu: ok then let's leave it as it is

src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/controllers/StaleBlobControllerTest.java
```diff
@@ -82,6 +84,21 @@ void should_return_list_of_stale_blobs_when_there_is_by_default_param_value() th
 
     }
 
+    @Test
+    void should_return_empty_data_when_there_is_no_stale_blob() throws Exception {
+
+        given(staleBlobFinder.findStaleBlobs(2)).willReturn(null);
```
rkondratowicz, line 90: shouldn't it return an empty list?

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/SearchResult.java
doncem: what buggers me more is that List<?> data can be null flushed

p.s. i accidentally pressed `unresolve` button

mdayican: normally `data` can not be null

src/main/java/uk/gov/hmcts/reform/blobrouter/util/TimeUtils.java
```diff
@@ -14,6 +14,8 @@ private TimeUtils() {
     }
 
     public static String toLocalTimeZone(Instant instant) {
-        return dateTimeFormatter.format(ZonedDateTime.ofInstant(instant, EUROPE_LONDON_ZONE_ID));
+        return instant == null
+            ? ""
```
rkondratowicz, line 18: why not null?

mdayican: envelope end point may return null

src/main/java/uk/gov/hmcts/reform/blobrouter/model/out/SearchResult.java
doncem: so do we then need to check for nulls? it's not normal as you pointed out

mdayican: I prefer to check.

PR #582, opened 2020-07-15
 add bulkscanauto containers to preview

mdayican:
### Change description ###

 adding preview containers

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/Chart.yaml
+++ b/charts/reform-scan-blob-router/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for Blob Router Service
 name: reform-scan-blob-router
 home: https://github.com/hmcts/blob-router-service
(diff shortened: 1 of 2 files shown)


mdayican:
> Could you please add proper description?
> 
> Description says bump chart version but also adding containers to preview osba.
> Was the version bump missed previously or is it done only to resolve sync issue ?
> 
> Changes look good but just thought would confirm above.

we do not need to bump version for preview changes main reason bump version.

nitinprabhu:
> > Could you please add proper description?
> > Description says bump chart version but also adding containers to preview osba.
> > Was the version bump missed previously or is it done only to resolve sync issue ?
> > Changes look good but just thought would confirm above.
> 
> we do not need to bump version for preview changes main reason bump version.

Thanks could you please remove adding of containers just to make it clear and not mix two changes in one pr ?

PR #587, opened 2020-07-17
Check status of the envelope before processing

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1321 (Blob router doesn't fetch envelope from DB after acquiring lease)

### Change description ###
Introduced predicate for checking if status of the envelope being processed is still CREATED and not changed due to any race condition.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
@@ -50,7 +50,7 @@ public class LeaseAcquirer {
 
     public void ifAcquiredOrElse(
         BlobClient blobClient,
-        Predicate<BlobClient> condition,
+        Predicate<BlobClient> blobCondition,
         Consumer<String> onSuccess,
         Consumer<BlobErrorCode> onFailure,
         boolean releaseLease
@@ -59,7 +59,7 @@ public class LeaseAcquirer {
             var leaseClient = leaseClientProvider.get(blobClient);
             var leaseId = leaseClient.acquireLease(LEASE_DURATION_IN_SECONDS);
 
-            if (condition.test(blobClient)) {
+            if (blobCondition.test(blobClient)) {
                 onSuccess.accept(leaseId);
             }
 
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
@@ -8,6 +8,7 @@ import org.slf4j.Logger;
 import org.springframework.stereotype.Component;
 
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -78,7 +81,11 @@ private void processBlob(BlobClient blob) {
             .ifPresentOrElse(
                 envelope -> {
                     if (envelope.status == Status.CREATED) {
-                        leaseAndThen(blob, () -> blobProcessor.continueProcessing(envelope.id, blob));
+                        leaseAndThen(
+                            blob,
+                            this::isEnvelopeInCreatedStatus,
```
lgonczar, line 86: Why are we involving this check in lease acquiring process?

lgonczar: Also, blobProcessor.process will need to do some custom stuff, too:
- create the envelope and process the blob when there's no envelope
- pick up the existing envelope if it's in the created state
- abandon processing if the envelope is not in created state

apereverzin: All this is done in LeaseAcquirer. And the check is called not while acquiring the lease, but after that.

src/test/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessorTest.java
```diff
@@ -140,10 +140,10 @@ private void dbHas(Envelope envelope) {
     @SuppressWarnings("unchecked")
     private void leaseCanBeAcquired() {
         doAnswer(invocation -> {
-            var okAction = (Consumer) invocation.getArgument(1);
+            var okAction = (Consumer) invocation.getArgument(2);
             okAction.accept(UUID.randomUUID().toString());
             return null;
-        }).when(leaseAcquirer).ifAcquiredOrElse(any(), any(), any(), anyBoolean());
+        }).when(leaseAcquirer).ifAcquiredOrElse(any(), any(), any(), any(), anyBoolean());
```
rkondratowicz, line 146: missing new test

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/LeaseAcquirer.java
```diff
@@ -37,12 +38,30 @@ public void ifAcquiredOrElse(
...
+            blob -> true,
+            onSuccess,
+            onFailure,
+            releaseLease
+        );
+    }
+
+    public void ifAcquiredOrElse(
+        BlobClient blobClient,
+        Predicate<BlobClient> blobCondition,
```
lgonczar, line 53: I disagree with this change - it's unnecessary and adds more complexity to the code that should only deal with acquiring leases. Why not just have this condition in the actual business logic?

apereverzin:
https://github.com/hmcts/blob-router-service/pull/589

PR #589, opened 2020-07-20
Added status check

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1321

### Change description ###
Added status check of the envelope before continuing processing

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
@@ -6,12 +6,14 @@ import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.models.BlobItem;
 import org.slf4j.Logger;
 import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.blobrouter.data.envelopes.Envelope;
 import uk.gov.hmcts.reform.blobrouter.data.envelopes.Status;
 import uk.gov.hmcts.reform.blobrouter.services.BlobReadinessChecker;
 import uk.gov.hmcts.reform.blobrouter.services.EnvelopeService;
 import uk.gov.hmcts.reform.blobrouter.services.storage.LeaseAcquirer;
 
 import java.time.Instant;
+import java.util.Optional;
 
 import static org.slf4j.LoggerFactory.getLogger;
 
@@ -73,18 +75,34 @@ public class ContainerProcessor {
     }
 
     private void processBlob(BlobClient blob) {
-        envelopeService
-            .findLastEnvelope(blob.getBlobName(), blob.getContainerName())
+        getLastEnvelope(blob)
             .ifPresentOrElse(
                 envelope -> {
                     if (envelope.status == Status.CREATED) {
-                        leaseAndThen(blob, () -> blobProcessor.continueProcessing(envelope.id, blob));
+                        leaseAndThen(blob, () -> {
+                            Optional<Envelope> envelopeOpt = getLastEnvelope(blob);
+                            if (envelopeOpt.isPresent()) {
+                                if (envelopeOpt.get().status == Status.CREATED) {
+                                    blobProcessor.continueProcessing(envelope.id, blob);
+                                } else {
+                                    logger.info(
+                                        "Envelope status changed in system, skipping. {} ",
+                                        envelope.getBasicInfo()
+                                    );
+                                }
+                            }
+                        });
                     } else {
                         logger.info("Envelope already processed in system, skipping. {} ", envelope.getBasicInfo());
                     }
                 },
                 () -> leaseAndThen(blob, () -> blobProcessor.process(blob))
             );
+    }
+
+    private Optional<Envelope> getLastEnvelope(BlobClient blob) {
+        return envelopeService
+            .findLastEnvelope(blob.getBlobName(), blob.getContainerName());
     }
 
     private void leaseAndThen(BlobClient blobClient, Runnable action) {
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
@@ -81,12 +81,9 @@ public class ContainerProcessor {
                 envelope -> {
                     if (envelope.status == Status.CREATED) {
                         leaseAndThen(blob, () ->
-                            handlePotentiallyChangedEnvelope(
+                            handlePotentiallyNonExistingEnvelope(
                                 blob,
-                                blobClient -> logger.error(
-                                    "Envelope no longer exists in system for blob {}, container {}",
-                                    blobClient.getBlobName(), blobClient.getContainerName()
-                                )
+                                this::logEnvelopeAlreadyDeleted
                             )
                         );
                     } else {
@@ -94,7 +91,7 @@ public class ContainerProcessor {
                     }
                 },
                 () -> leaseAndThen(blob, () ->
-                    handlePotentiallyChangedEnvelope(
+                    handlePotentiallyNonExistingEnvelope(
                         blob,
                         blobProcessor::process
                     )
@@ -102,11 +99,14 @@ public class ContainerProcessor {
             );
     }
 
-    private void handlePotentiallyChangedEnvelope(BlobClient blob, Consumer<BlobClient> notPresentEnvelopeHandler) {
+    private void handlePotentiallyNonExistingEnvelope(
+        BlobClient blob,
+        Consumer<BlobClient> nonExistingEnvelopeHandler
+    ) {
         getLastEnvelope(blob)
             .ifPresentOrElse(
                 envelopePotentiallyProcessed -> continueProcessingIfPossible(blob, envelopePotentiallyProcessed),
-                () -> notPresentEnvelopeHandler.accept(blob)
+                () -> nonExistingEnvelopeHandler.accept(blob)
             );
     }
 
@@ -118,6 +118,13 @@ public class ContainerProcessor {
         }
     }
 
+    private void logEnvelopeAlreadyDeleted(BlobClient blob) {
+        logger.error(
+            "Envelope already deleted in system for blob {}, container {}",
+            blob.getBlobName(), blob.getContainerName()
+        );
+    }
+
     private void logEnvelopeAlreadyProcessed(Envelope envelope) {
         logger.info("Envelope already processed in system, skipping. {} ", envelope.getBasicInfo());
     }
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,12 +75,23 @@ private boolean isReady(BlobItem blob, String containerName) {
...
+                            if (envelopeOpt.isPresent()) {
+                                if (envelopeOpt.get().status == Status.CREATED) {
+                                    blobProcessor.continueProcessing(envelope.id, blob);
+                                } else {
+                                    logger.info(
+                                        "Envelope status changed in system, skipping. {} ",
+                                        envelope.getBasicInfo()
+                                    );
+                                }
+                            }
```
lgonczar, line 93: We should log an error in `else` if the envelope doesn't exist.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,12 +75,23 @@ private boolean isReady(BlobItem blob, String containerName) {
...
-        envelopeService
-            .findLastEnvelope(blob.getBlobName(), blob.getContainerName())
+        getLastEnvelope(blob)
             .ifPresentOrElse(
                 envelope -> {
                     if (envelope.status == Status.CREATED) {
-                        leaseAndThen(blob, () -> blobProcessor.continueProcessing(envelope.id, blob));
+                        leaseAndThen(blob, () -> {
+                            Optional<Envelope> envelopeOpt = getLastEnvelope(blob);
+                            if (envelopeOpt.isPresent()) {
```
lgonczar, line 84: We can use `ifPresentOrElse`, so that we won't have to execute `get` on the optional in the body.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,12 +75,23 @@ private boolean isReady(BlobItem blob, String containerName) {
...
+                                if (envelopeOpt.get().status == Status.CREATED) {
+                                    blobProcessor.continueProcessing(envelope.id, blob);
+                                } else {
+                                    logger.info(
+                                        "Envelope status changed in system, skipping. {} ",
```
lgonczar, line 89: Best to have the same message as below, i.e. that the envelope has already been processed in the system. That's the information we care about the most. The fact that its status has changed since we acquired the lease, not so much. We can mention that, though.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,12 +75,23 @@ private boolean isReady(BlobItem blob, String containerName) {
     }
 
     private void processBlob(BlobClient blob) {
-        envelopeService
-            .findLastEnvelope(blob.getBlobName(), blob.getContainerName())
+        getLastEnvelope(blob)
             .ifPresentOrElse(
```
lgonczar, line 79: We also need this additional envelope fetch in the else clause, as the envelope might have been created.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,20 +75,43 @@ private boolean isReady(BlobItem blob, String containerName) {
...
+                                    },
+                                    () -> logger.error(
+                                        "Envelope no longer exists in system for blob {}, container {}",
+                                        blob.getBlobName(), blob.getContainerName()
+                                    )
+                                );
+                        });
                     } else {
-                        logger.info("Envelope already processed in system, skipping. {} ", envelope.getBasicInfo());
+                        logProcessedEnvelope(envelope);
```
doncem: here's a wild thought: how about we try lease straight away before making db call? acquirer already checks `exc.getErrorCode() != LEASE_ALREADY_PRESENT && exc.getErrorCode() != BLOB_NOT_FOUND` and doesn't log an error if this condition is not satisfied. which is what we need right? then on else part we'll just log info as we do now.

by the looks here we just overusing if/else/option/if/else things. plus multiple db calls.

in case got rejected? we have duplication covered. normally such situation (file has already been processed) should not happen. so either we doing it wrong or orchestration around here with all these libraries and services doing their own thing is incorrect.

is this just a wild thought or has some value? if no other way - i'm ok to participate in chain of approvals required

apereverzin: I thought about acquiring lease straight away but my concern is about performance - if several pods will be acquiring - releasing the lease of the file this can make performance impact, bigger than querying the db

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,28 +76,72 @@ private boolean isReady(BlobItem blob, String containerName) {
...
     private void processBlob(BlobClient blob) {
-        envelopeService
-            .findLastEnvelope(blob.getBlobName(), blob.getContainerName())
+        getLastEnvelope(blob)
             .ifPresentOrElse(
                 envelope -> {
                     if (envelope.status == Status.CREATED) {
-                        leaseAndThen(blob, () -> blobProcessor.continueProcessing(envelope.id, blob));
+                        leaseAndThen(blob, () ->
+                            handlePotentiallyNonExistingEnvelope(
```
rkondratowicz, line 84: can it not existi if it's status is 'created'?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,28 +76,72 @@ private boolean isReady(BlobItem blob, String containerName) {
...
+                    handlePotentiallyNonExistingEnvelope(
+                        blob,
+                        blobProcessor::process
+                    )
+                )
             );
     }
 
-    private void leaseAndThen(BlobClient blobClient, Runnable action) {
+    private void handlePotentiallyNonExistingEnvelope(
```
rkondratowicz, line 102: method name doesn't say that it continues processing the envelope

apereverzin:
Renamed to continueProcessingEnvelopeIfItStillExists

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,28 +76,72 @@ private boolean isReady(BlobItem blob, String containerName) {
...
     private void processBlob(BlobClient blob) {
-        envelopeService
-            .findLastEnvelope(blob.getBlobName(), blob.getContainerName())
+        getLastEnvelope(blob)
             .ifPresentOrElse(
                 envelope -> {
                     if (envelope.status == Status.CREATED) {
-                        leaseAndThen(blob, () -> blobProcessor.continueProcessing(envelope.id, blob));
+                        leaseAndThen(blob, () ->
+                            continueProcessingEnvelopeIfItStillExists(
```
lgonczar, line 84: How about `continueProcessingEnvelopeIfEligible` or sth like this? Based on the code, existing alone is not sufficient for an envelope to be processed.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,28 +76,72 @@ private boolean isReady(BlobItem blob, String containerName) {
...
+            logEnvelopeAlreadyProcessed(envelope);
+        }
+    }
+
+    private void logEnvelopeAlreadyDeleted(BlobClient blob) {
+        logger.error(
+            "Envelope already deleted in system for blob {}, container {}",
+            blob.getBlobName(), blob.getContainerName()
+        );
+    }
```
lgonczar, line 126: Suggestion: maybe it would be best to get rid of `already` from the names and message? We typically use already when there's no need to do something, because it has already happened. This, however, is an error scenario -envelopes should never be deleted.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,28 +76,72 @@ private boolean isReady(BlobItem blob, String containerName) {
...
         leaseAcquirer.ifAcquiredOrElse(
-            blobClient,
+            blob,
             leaseId -> action.run(),
             errorCode -> logger.info(
                 "Cannot acquire a lease for blob - skipping. File name: {}, container: {}, error code: {}",
-                blobClient.getBlobName(),
-                blobClient.getContainerName(),
+                blob.getBlobName(),
+                blob.getContainerName(),
```
lgonczar, line 144: Why are we renaming it to `blob` if it's a client?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,28 +76,72 @@ private boolean isReady(BlobItem blob, String containerName) {
...
     }
 
-    private void leaseAndThen(BlobClient blobClient, Runnable action) {
+    private void continueProcessingEnvelopeIfItStillExists(
+        BlobClient blob,
+        Consumer<BlobClient> nonExistingEnvelopeHandler
+    ) {
+        getLastEnvelope(blob)
+            .ifPresentOrElse(
+                envelopePotentiallyProcessed -> continueProcessingIfPossible(blob, envelopePotentiallyProcessed),
```
lgonczar, line 108: How about just `envelope`?

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -73,28 +76,72 @@ private boolean isReady(BlobItem blob, String containerName) {
...
+                            )
+                        );
                     } else {
-                        logger.info("Envelope already processed in system, skipping. {} ", envelope.getBasicInfo());
+                        logEnvelopeAlreadyProcessed(envelope);
                     }
                 },
-                () -> leaseAndThen(blob, () -> blobProcessor.process(blob))
+                () -> leaseAndThen(blob, () ->
+                    continueProcessingEnvelopeIfItStillExists(
```
lgonczar, line 94: The name of this method is invalid in this context, because we're in a situation where the envelope initially didn't exist.

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -58,35 +61,79 @@ public void process(String containerName) {
...
+
+    private void logEnvelopeDeleted(BlobClient blobClient) {
+        logger.error(
+            "Envelope deleted in system for blob {}, container {}",
+            blobClient.getBlobName(), blobClient.getContainerName()
+        );
+    }
+
+    private void logEnvelopeProcessed(Envelope envelope) {
+        logger.info("Envelope processed in system, skipping. {} ", envelope.getBasicInfo());
```
lgonczar, line 129: `already` was actually good here and in the methods's name - it adds useful information. It was just misleading for deletion, as deletion should never happen.

PR #595, opened 2020-07-30
Use new method signature as jenkins library is updated

nitinprabhu:


nitinprabhu:
Merging this one as https://github.com/hmcts/blob-router-service/pull/595 will fix the build but suppression needs to go first

PR #613, opened 2020-08-14
Add config to enable/disable reconciliation

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1359

### Change description ###
Add config to enable or disable reconciliation endpoint.
Fluc PR: https://github.com/hmcts/cnp-flux-config/pull/5462

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.yaml
+++ b/charts/reform-scan-blob-router/values.yaml
@@ -44,6 +44,7 @@ java:
 
     SMTP_HOST: "false"
     SEND_DAILY_REPORT_ENABLED: "false"
+    RECONCILIATION_ENABLED: "true"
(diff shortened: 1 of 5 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -44,6 +44,7 @@ java:
 
     SMTP_HOST: "false"
     SEND_DAILY_REPORT_ENABLED: "false"
+    RECONCILIATION_ENABLED: "true"
```
nitinprabhu, line 47: may be safer side let's keep it disabled and only enable in aat template ?
Once auth story is played this will be removed completely so you won't need to bother with flux PR's

aliveni: Updated the templates

PR #626, opened 2020-08-24
Update Swagger configuration to publish reconciliation api

aliveni:
### Change description ###
Swagger is not publishing Reconciliation API because the controller is in `uk.gov.hmcts.reform.blobrouter.reconciliation.controller`. 

see https://hmcts.github.io/reform-api-docs/specs/blob-router-service.json

```
[ ] Yes
[x] No
```


doncem (approved):
Interesting how this is different. Normally microservice would provide as minimum as possible. Including set of controllers defined in certain package.

But yea. NVM

nitinprabhu (approved):

PR #629, opened 2020-08-26
Add reconciliation endpoint to api mgmt

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1366

```
[ ] Yes
[x] No
```


infrastructure/api-mgmt.tf
```diff
@@ -46,3 +46,27 @@ module "api_mgmt_policy" {
...
+
+module "api_mgmt_reconciliation" {
+  source        = "git@github.com:hmcts/cnp-module-api-mgmt-api?ref=master"
+  name          = "${var.component}-reconciliation-api"
+  api_mgmt_name = "core-api-mgmt-${var.env}"
+  api_mgmt_rg   = "core-infra-${var.env}"
+  revision      = "1"
+  product_id    = "${module.api_mgmt_product.product_id}"
+  display_name  = "Blob Router Reconciliation API"
+  path          = "reform-scan/reconciliation-report"
```
doncem, line 58: i think this will mean that the full endpoint for api request will be `{core-api-url}/reform-scan/reconciliation-report/reform-scan-reconciliation-report/{date}`

perhaps just `path = "reconciliation"` is enough?

doncem: p.s. mind you, all endpoints will be available what's present in current swagger spec at the time of infra execution

src/smokeTest/java/uk/gov/hmcts/reform/blobrouter/ReconciliationApiTest.java
```diff
@@ -0,0 +1,98 @@
...
+        validApiKey = getValidReconciliationApiKey();
+    }
+
+    @Test
+    void should_accept_request_with_valid_api_key() {
+        Response response = callReconciliationEndpoint(validApiKey).thenReturn();
+
+        assertThat(response.getStatusCode()).isEqualTo(OK.value());
+        assertThat(response.body().jsonPath().getString("id")).isNotEmpty();
+    }
```
doncem, line 44: we should not post anything which ends up in success. smoke tests are executed on prod environment as well

doncem (commented):
can we also include synthetic header for smoke tests too :)

infrastructure/api-mgmt.tf
```diff
@@ -46,3 +46,27 @@ module "api_mgmt_policy" {
...
+  protocols = [
+    "https"
+  ]
+  service_url = "http://${var.product}-${var.component}-${var.env}.service.core-compute-${var.env}.internal"
+  swagger_url = "https://hmcts.github.io/reform-api-docs/specs/blob-router-service.json"
```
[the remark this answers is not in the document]
aliveni: Raised https://github.com/hmcts/blob-router-service/pull/630

src/smokeTest/java/uk/gov/hmcts/reform/blobrouter/ReconciliationApiTest.java
aliveni: how can we test the valid scenario? is it possible?

doncem: only through functional test. it is a POST endpoint - state changer in actual API. smoke test should not do it at any point in future

infrastructure/api-mgmt.tf
```diff
@@ -46,3 +46,27 @@ module "api_mgmt_policy" {
...
+
+module "api_mgmt_reconciliation" {
+  source        = "git@github.com:hmcts/cnp-module-api-mgmt-api?ref=master"
+  name          = "${var.component}-reconciliation-api"
+  api_mgmt_name = "core-api-mgmt-${var.env}"
+  api_mgmt_rg   = "core-infra-${var.env}"
+  revision      = "1"
+  product_id    = "${module.api_mgmt_product.product_id}"
+  display_name  = "Blob Router Reconciliation API"
+  path          = "reconciliation"
```
doncem, line 58: would be nice to re-use same path - `reform-scan`. i believe name of module is what important - not the path. then we can have actual endpoint in the shape we actually want: `/reform-scan/reconciliation-report/{date}`

suggesting that because service endpoint is now better: `/reconciliation-report/{date}` ;)

doncem: p.s. perhaps later on both APIs can be merged and marked as `revision = 2`. policy changes are in question here - not sure how to adapt to different endpoints but there must be some sort of filter

doncem (approved):

aliveni:
Tested API management changes in demo. I added a query param to the swagger spec URL and renamed the policy file name to apply the changes.

infrastructure/reconciliation-api-policy.xml
```diff
@@ -0,0 +1,29 @@
+<policies>
```
doncem, line 1: where is this file referenced?

aliveni: It's no longer used. I'll remove it

aliveni: Removed the unused policy file

mdayican (approved):

PR #633, opened 2020-08-26
fix pipeline - change docker container in test and disable smoke test

mdayican:
### Change description ###

Pipeline broken as test container creation changed.
Latest image does not support `copyFromUrl`.
what is wrong with it but always we have problem with it :)

And disable smoke test as Test cert is corrupted. will be fixed in another PR.

```
[ ] Yes
[ ] No
```


doncem (approved):

aliveni (approved):

nitinprabhu (approved):

nitinprabhu (changes requested):
can we plz extract disabling of the test as a separate Pr ?

nitinprabhu (approved):

PR #638, opened 2020-09-02
Update Unauthorized error response in Reconciliation endpoint

aliveni:
### Change description ###
When the authorization header is invalid, the Reconciliation endpoint returns the response with the error message and cause. 
Updated the Exception handler to return the error message only.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/exceptionhandler/ResponseExceptionHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/exceptionhandler/ResponseExceptionHandler.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.reform.blobrouter.exceptionhandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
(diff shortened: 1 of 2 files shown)


mdayican (approved):

nitinprabhu (approved):

src/main/java/uk/gov/hmcts/reform/blobrouter/exceptionhandler/ResponseExceptionHandler.java
```diff
@@ -51,10 +52,9 @@ protected ErrorResponse handleInvalidSupplierStatementException(InvalidSupplierS
     }
 
     @ExceptionHandler(InvalidApiKeyException.class)
-    @ResponseStatus(HttpStatus.UNAUTHORIZED)
-    protected ErrorResponse handleInvalidApiKeyException(InvalidApiKeyException exception) {
+    protected ResponseEntity<String> handleInvalidApiKeyException(InvalidApiKeyException exception) {
         log.error(exception.getMessage(), exception);
-        return new ErrorResponse(exception.getMessage(), exception.getClass());
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(exception.getMessage());
```
rkondratowicz, line 57: Now this is the only error that results in a different payload.

aliveni: what do you suggest? shall I leave it as it is now?

rkondratowicz: Yeah, imho it's better to keep it consistent.
If we want to change it, let's change it for all errors.

rkondratowicz: Also probably better to return json rather than just text

aliveni:
I'll raise a new PR to remove the `clause` from all instances in the Exception Handler class

PR #653, opened 2020-09-07
Add PCQ storage configuration

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1377

### Change description ###
- Add PCQ backend URL to retrieve the SAS token
- Add PCQ storage url
- Copy blob router s2s secret from s2s key vault to reform-scan key vault

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.aat.template.yaml
+++ b/charts/reform-scan-blob-router/values.aat.template.yaml
@@ -22,8 +22,10 @@ java:
         - storage-account-staging-name
         - storage-account-staging-primary-key
         - storage-account-staging-secondary-key
+        - pcq-storage-account-staging-name
+        - pcq-storage-account-staging-primary-key
+        - pcq-storage-account-staging-secondary-key
         - crime-storage-connection-string
-        - pcq-storage-connection-string
         - blob-router-staging-db-password
         - blob-router-staging-db-user
         - blob-router-staging-db-host
--- a/charts/reform-scan-blob-router/values.aat.template.yaml
+++ b/charts/reform-scan-blob-router/values.aat.template.yaml
@@ -1,6 +1,7 @@
 java:
   environment:
     CRIME_ENABLED: "true"
+    PCQ_ENABLED: "true"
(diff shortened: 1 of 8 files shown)


charts/reform-scan-blob-router/values.aat.template.yaml
```diff
@@ -22,6 +23,7 @@ java:
         - storage-account-staging-primary-key
         - storage-account-staging-secondary-key
         - crime-storage-connection-string
+        - pcq-storage-connection-string
```
doncem, line 26: given exists in required environments (including prod) - all good

doncem (approved):

charts/reform-scan-blob-router/values.aat.template.yaml
aliveni: I added it in aat. The storage account doesn't exist in other environments. I'll add the secret to other environments with some value before I merge the PR

doncem: if values gonna be used in any environment app - they must exist. e.g. storage clients might be ok until we make a call. but building clients might involve some validation. plus pods won't boot up - it'll error saying secret not found

rkondratowicz: story description says SAS token should be used for uploading files

aliveni: I'll update the PR

aliveni: Added pcq storage account name and account keys. Added the secrets in AAT. I'll add the secrets in other environments before merging the PR.
Removed connection string and deleted the secret from aat key vault.

mdayican (approved):

doncem (approved):

charts/reform-scan-blob-router/values.aat.template.yaml
aliveni: As discussed, removed all secrets related to pcq storage account and just added storage url.

PR #660, opened 2020-09-09
add summary task

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1361

### Change description ###

add summary task

```
[ ] Yes
[x] No
```


doncem (commented):
not too big so maybe ok but i think we prefer config first - then code using it

charts/reform-scan-blob-router/values.yaml
```diff
@@ -39,6 +39,9 @@ java:
     SEND_DAILY_REPORT_CRON: "0 0 6 ? * *"
     SEND_NOTIFICATIONS_CRON: "0 0/10 * * * *"
 
+    CREATE_SUMMARY_REPORT_CRON: "0 0/30 1-11 * * * "
+    CREATE_SUMMARY_REPORT_ENABLED: true
```
rkondratowicz, line 43: are you sure we want to enable it now?

charts/reform-scan-blob-router/values.yaml
```diff
@@ -39,6 +39,9 @@ java:
     SEND_DAILY_REPORT_CRON: "0 0 6 ? * *"
     SEND_NOTIFICATIONS_CRON: "0 0/10 * * * *"
 
+    CREATE_SUMMARY_REPORT_CRON: "0 0/30 1-11 * * * "
```
rkondratowicz, line 42: will this run at 11:30? Story description says last time it should run is 11:00

src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/SummaryReportTask.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.time.LocalDate;
+
+import static java.time.ZoneOffset.UTC;
+import static org.slf4j.LoggerFactory.getLogger;
+import static uk.gov.hmcts.reform.blobrouter.util.TimeZones.EUROPE_LONDON;
+
+@Component
+@ConditionalOnProperty(value = "scheduling.task.create-summary-report.enabled")
+public class SummaryReportTask {
```
rkondratowicz, line 18: it's worth clarifying it's about reconciliation. Either in class name or package

src/main/resources/application.yaml
```diff
@@ -132,6 +132,9 @@ scheduling:
     send-daily-report:
       cron: ${SEND_DAILY_REPORT_CRON}
       enabled: ${SEND_DAILY_REPORT_ENABLED}
+    create-summary-report:
+      cron: ${CREATE_SUMMARY_REPORT_CRON}
+      enabled: ${CREATE_SUMMARY_REPORT_ENABLED}
```
rkondratowicz, line 137: make it clear it's about reconciliation

doncem (approved):

apereverzin (approved):

PR #670, opened 2020-09-11
Add pcq storage url config

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1377

### Change description ###
Add PCQ storage URL config

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.preview.template.yaml
+++ b/charts/reform-scan-blob-router/values.preview.template.yaml
@@ -46,9 +46,9 @@ java:
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
 
     STORAGE_BULKSCAN_URL: https://bulkscan.aat.platform.hmcts.net
+    STORAGE_PCQ_URL: "https://pcqsharedaat.blob.core.windows.net"
 
     STORAGE_URL: "https://$(STORAGE_ACCOUNT_NAME).blob.core.windows.net"
-    STORAGE_PCQ_URL: "https://$(STORAGE_PCQ_ACCOUNT_NAME).blob.core.windows.net"
     STORAGE_CRIME_CONNECTION_STRING: "DefaultEndpointsProtocol=https;AccountName=$(STORAGE_CRIME_ACCOUNT_NAME);AccountKey=$(STORAGE_CRIME_ACCOUNT_KEY);EndpointSuffix=core.windows.net"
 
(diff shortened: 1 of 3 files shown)


charts/reform-scan-blob-router/values.yaml
```diff
@@ -46,6 +46,8 @@ java:
 
     STORAGE_URL: https://reformscan.{{ .Values.global.environment }}.platform.hmcts.net
     STORAGE_BULKSCAN_URL: https://bulkscan.{{ .Values.global.environment }}.platform.hmcts.net
+    STORAGE_PCQ_URL: https://pcqshared{{ .Values.global.environment }}.blob.core.windows.net
```
doncem, line 49: are we sure it'll have all these names?

aliveni: at the moment, storage account exists in aat only but it will be created with the same format

rkondratowicz (approved):

charts/reform-scan-blob-router/values.yaml
doncem: need to be sure config is not used. well, not like used but it won't make any calls or smth. otherwise it'll fail

doncem (approved):

charts/reform-scan-blob-router/values.yaml
aliveni: pcq container is disabled so it'll not be used

aliveni: It'll be called only when dispatching the pcq envelopes

charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -46,9 +46,9 @@ java:
     STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
 
     STORAGE_BULKSCAN_URL: https://bulkscan.aat.platform.hmcts.net
+    STORAGE_PCQ_URL: "https://pcqsharedaat.blob.core.windows.net"
```
mdayican, line 49: this does not seems correct to me.
is should be osba??

aliveni: we've been using aat storage account for bulkscan containers. wouldn't it be the same for pcq?

mdayican (approved):

PR #684, opened 2020-09-16
add mail and mail sending task config

mdayican:
https://tools.hmcts.net/jira/browse/BPS-1369

### Change description ###

add mail and mail sending task config.
using existing report recipients and mail-from.

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -146,6 +149,9 @@ proxy:
 
 reconciliation:
   api-key: ${RECONCILIATION_API_KEY}
+  report:
+      mail-from: ${SMTP_USERNAME}
+      mail-recipients: ${REPORTS_RECIPIENTS}
```
mdayican, line 154: using same values like other reports

doncem: i guess we want to keep it separate in case recipients need to be different

doncem (approved):

src/integrationTest/resources/application.properties
```diff
@@ -47,6 +47,9 @@ scheduling.task.create-reconciliation-summary-report.cron=0/10 * * * * *
 scheduling.task.create-reconciliation-detailed-report.enabled=false
 scheduling.task.create-reconciliation-detailed-report.cron=0/10 * * * * *
 
+scheduling.task.send-reconciliation-report-mail.enabled=false
+scheduling.task.send-reconciliation-report-mail.cron=0/10 * * * * *
```
rkondratowicz, line 51: every 10 seconds?

rkondratowicz (approved):

src/main/resources/application.yaml
mdayican: that is why I created different properties for now.

PR #707, opened 2020-09-25
Refactor Reconciliation Mail Service to generate and email reports manually

aliveni:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-1399

### Change description ###
Add `automatedEmail` boolean flag to ReconcilitionMailService methods to skip the report already sent condition.
When manually triggered to generate and send reports and the automated email was sent already, it should generate and email the reports again.
This will be used in the next PR.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #708, opened 2020-09-25
Refactor Reconciliation Mail Service to generate and email reports manually

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1399

### Change description ###
Duplicate of #707 
Add `automatedEmail` boolean flag to ReconcilitionMailService methods to skip the report already sent condition.
When manually triggered to generate and send reports and the automated email was sent already, it should generate and email the reports again.
This will be used in the next PR.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/reconciliation/service/ReconciliationMailService.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/reconciliation/service/ReconciliationMailService.java
@@ -63,14 +63,14 @@ public class ReconciliationMailService {
         this.mailRecipients = recipients;
     }
 
-    public void process(LocalDate date, List<TargetStorageAccount> availableAccounts) {
+    public void process(LocalDate date, List<TargetStorageAccount> availableAccounts, boolean automatedEmail) {
         Optional<EnvelopeSupplierStatement> optionalEnvelopeSupplierStatement = supplierStatementRepository
             .findLatest(date);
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/reconciliation/service/ReconciliationMailService.java
```diff
@@ -81,25 +81,27 @@ public void process(LocalDate date, List<TargetStorageAccount> availableAccounts
...
         ReconciliationReport reconciliationReport,
-        TargetStorageAccount account
+        TargetStorageAccount account,
+        boolean automatedEmail
     ) {
         try {
             if (reconciliationReport.summaryContent != null) {
 
-                if (reconciliationReport.sentAt != null) {
+                if (automatedEmail && reconciliationReport.sentAt != null) {
```
rkondratowicz, line 104: so it stops here never sends an email if it's automated?

rkondratowicz:
Will this actually be needed if a new report gets generated?
This change makes it possible to force resending the same report again (which I think is not what we need)

aliveni:
Closing the PR as the build is failing.

aliveni:
> Will this actually be needed if a new report gets generated?
> This change makes it possible to force resending the same report again (which I think is not what we need)

As discussed, this PR is not required. We send the email only when there is a new report generated.

PR #711, opened 2020-09-28
use lower case to compare

mdayican:
### Change description ###

use lowercase to find container mapping.
provider can send container name in capital letters. 
May be we should reject the supplier statement if container name does not exactly match?

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/blobrouter/reconciliation/service/SummaryReportServiceTest.java
```diff
@@ -248,7 +248,7 @@ void should_save_reports_if_there_is_supplier_report() throws IOException, SQLEx
 
         List envelopeList = Arrays.asList(
             createEnvelope("1010404021234_14-08-2020-08-31.zip", "probate"),
-            createEnvelope("9810404021234_14-08-2020-03-08-31.zip", "sscs"),
+            createEnvelope("9810404021234_14-08-2020-03-08-31.zip", "SSCS"),
```
rkondratowicz, line 251: please add a comment in code here to clarify why it's uppercase

PR #721, opened 2020-09-29
Endpoint to generate and email Reconciliation reports

aliveni:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-1399

### Change description ###
Re-opening #720 
Added endpoint with the path `reconciliation-reports/generate-and-email-report` 
- Generates reconciliation reports both summary and detailed summary reports.
- Sends email when new reports are generated.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #724, opened 2020-09-30
Add configuration for Mail sender when email config is disabled

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1399

### Change description ###
This configuration will be used in #721 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/config/EmailConfig.java
```diff
@@ -0,0 +1,33 @@
...
+        return new MessageSender() {
+            @Override
+            public void sendMessageWithAttachments(
+                String subject,
+                String body,
+                String from,
+                String[] recipients,
+                Map<String, File> attachments
+            ) {
+                // do nothing
```
akomisarek, line 28: Isn't only for dev purposes? If used maybe would be beneficial to log something? What do you think?

aliveni: I agree. added log

aliveni:
Closing the PR as the preview build is failing and unable to see the error because the container is not created

PR #725, opened 2020-10-01
Add configuration for Mail sender when email config is disabled

aliveni:
### Change description ###
Re-opening #724 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/config/EmailConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Lazy;
+import uk.gov.hmcts.reform.blobrouter.services.email.MessageSender;
+
+import java.io.File;
+import java.util.Map;
+
+@Configuration
+@Lazy
```
mdayican, line 15: why lazy?

aliveni: It'll create the bean only when required

mdayican: if there is any error I think it is better to see them in loading .

mdayican (approved):

PR #730, opened 2020-10-05
add report date to report table

mdayican:
### Change description ###

add report date to report table.
currently we use create_at which is not clear. 
Report date and creation might be different.  Now there is 'default' it will be 'not null' in second version with code changes.
name suggestions welcomed. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/db/migration/V017__add_new_column_reconciliation_report.sql
+++ b/src/main/resources/db/migration/V017__add_new_column_reconciliation_report.sql
@@ -1,0 +1,2 @@
+ALTER TABLE envelope_reconciliation_reports
+ADD COLUMN report_for DATE DEFAULT NOW();


src/main/resources/db/migration/V017__add_new_column_reconciliation_report.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE envelope_reconciliation_reports
+ADD COLUMN report_for DATE DEFAULT NOW();
```
rkondratowicz, line 2: do we need it or can use use date from related statement?

mdayican: mail service does not know supplier statement. sends the latest report and assume create_at is same as report_for date

rkondratowicz: it doesn't need to.
fix the SQL query in `findByDate` repository method to use the statement column.

rkondratowicz (changes requested):
no need, just fix the query for finding reports

PR #732, opened 2020-10-05
fix: bump testcontainers to fix 2.4.0 docker issue

akomisarek:
### Change description ###

https://github.com/testcontainers/testcontainers-java/pull/3159 fixed in new testcontainers release candidate:
https://github.com/testcontainers/testcontainers-java/releases/tag/1.15.0-rc2

I would believe we should upgrade, it's only test dependency and if it fixes the problem, it's better than downgrading docker doing the workarounds on local.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeControllerTest.java
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.reform.blobrouter.controllers;
 
-import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/blobrouter/controllers/EnvelopeControllerTest.java
```diff
@@ -229,12 +227,6 @@ private Envelope envelope(String fileName, String container, Instant createdDate
         );
     }
 
-    @NotNull
-    private Envelope envelope(String fileName, String container) {
```
aliveni, line 233: why is this removed? Is it not used?

aliveni: Just checked the code. it is not used

aliveni (approved):

PR #738, opened 2020-10-07
set crons for reconciliation tasks

mdayican:
### Change description ###

set crons for reconciliation tasks.
what do you think ?

```
[ ] Yes
[x] No
```


charts/reform-scan-blob-router/values.yaml
```diff
@@ -39,11 +39,11 @@ java:
     SEND_DAILY_REPORT_CRON: "0 0 6 ? * *"
     SEND_NOTIFICATIONS_CRON: "0 0/10 * * * *"
 
-    CREATE_RECONCILIATION_SUMMARY_REPORT_CRON: "0 */30 1-11 * * *"
+    CREATE_RECONCILIATION_SUMMARY_REPORT_CRON: "0 0 1-6 * * *"
     CREATE_RECONCILIATION_SUMMARY_REPORT_ENABLED: "false"
-    CREATE_RECONCILIATION_DETAILED_REPORT_CRON: "0 */30 2-12 * * *"
+    CREATE_RECONCILIATION_DETAILED_REPORT_CRON: "0 0 3-7 * * *"
```
akomisarek, line 44: Could it start at 2? why the 1 hour delay?

akomisarek (approved):

charts/reform-scan-blob-router/values.yaml
mdayican: it can .

PR #839, opened 2020-12-15
set db product name

mdayican:
### Change description ###

Request to enable us to setup db access for reform scan

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -12,6 +12,7 @@ module "reform-blob-router-db" {
   source             = "git@github.com:hmcts/cnp-module-postgres?ref=master"
   product            = var.product
   component          = var.component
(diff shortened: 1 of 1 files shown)


infrastructure/main.tf
```diff
@@ -10,7 +10,8 @@ locals {
 
 module "reform-blob-router-db" {
   source             = "git@github.com:hmcts/cnp-module-postgres?ref=master"
-  product            = var.component
+  product            = var.product
+  component          = var.component
```
timja, line 14: the previous name was only using the component so we need to keep using it by providing the name override
```suggestion
  component          = var.component
  name               = var.component
```

timja (approved):

PR #884, opened 2021-01-25
Copy blob for cft

mdayican:
### Change description ###

Copy Async. blob for cft

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProxy.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProxy.java
@@ -23,7 +23,6 @@ import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
 import java.util.concurrent.TimeUnit;
 
-import static com.azure.storage.blob.models.CopyStatusType.PENDING;
 import static org.slf4j.LoggerFactory.getLogger;
 
 @Service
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProxy.java
```diff
@@ -93,4 +105,56 @@ public void upload(
...
+                );
+            final BlockBlobClient targetBlob =
+                get(targetStorageAccount, destinationContainer)
+                    .getBlobClient(blobName)
+                    .getBlockBlobClient();
+
+            var start = System.nanoTime();
+            final SyncPoller<BlobCopyInfo, Void> poller =
+                targetBlob.beginCopy(sourceBlob.getBlobUrl() + "?" + sasToken, Duration.ofSeconds(2));
+            PollResponse<BlobCopyInfo> pollResponse = poller.waitForCompletion(Duration.ofMinutes(10));
```
apereverzin, line 135: We should check status of PollResponse and continue depending on the result - SUCCESSFULLY_COMPLETED or FAILED.

mdayican: I though it may stay on pending status forever that is why I add timeout. Processor side is not doing anything if it the status is not SUCCESS.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
```diff
@@ -39,4 +40,26 @@ public void dispatch(
...
+        TargetStorageAccount targetStorageAccount
+    ) {
+        logger.info(
+            "Move file from: {} to Container: {}. Storage: {}",
+            blob.getBlobUrl(),
+            destinationContainer,
+            targetStorageAccount
+        );
+
+        blobContainerClientProxy.moveBlob(blob, destinationContainer, targetStorageAccount);
```
apereverzin, line 56: Or maybe check the status of PollResponse here, return the result in such a case.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobContainerClientProxy.java
apereverzin: Maybe there should be some retry policy? Anyway failure must be handled.

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
mdayican: I think we do not need dispatcher . I will tidy up the code, first need to find the way to upload for PCQ and crime. 
after that CFT file copy need to change as well as we do not need to transfer all zip file.

PR #929, opened 2021-02-18
Envelope action service

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-363

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/services/EnvelopeActionService.java
```diff
@@ -0,0 +1,87 @@
...
+        if (envelope.status == DISPATCHED || envelope.status == REJECTED
+            || !isStale(envelope)) {
+            throw new EnvelopeCompletedOrNotStaleException(
+                "Envelope with id " + envelope.id + " is completed or not stale"
+            );
+        }
+    }
+
+    private boolean isStale(Envelope envelope) {
+        if (envelope.status != Status.CREATED) {
```
mdayican, line 75: addition to this, if it is CFT envelope can we check processor?

apereverzin: Processor and router have different databases. As mentioned in https://tools.hmcts.net/jira/browse/DTSBPS-348?focusedCommentId=952256&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-952256 there are several zip files in router which are not known by processor. This means that we need to check and handle in processor separately.

mdayican: there is an end point to check

apereverzin: The idea here is to handle an envelope which is stale in router i.e. in CREATED status with after e.g. 2 hours. It does not matter if it exists in processor.

mdayican: if file is processed in processor successfully you can not mark as rejected?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/events/EventType.java
```diff
@@ -9,4 +9,6 @@
     DUPLICATE_REJECTED,
     ERROR,
     NOTIFICATION_SENT,
+    MANUALLY_MARKED_AS_REJECTED,
+    MANUALLY_MARKED_AS_DISPATCHED
```
mdayican, line 13: do we use event type in any report or would it affect any reports?

apereverzin: In this report
image

mdayican: I mean if we are classifying with event type that would affect the report result.

mdayican (approved):

PR #930, opened 2021-02-18
refactor- move uploadwithchunks to blobmover

mdayican:
### Change description ###

move existing uploadwithchunks method to blobmover, new method is not in use now.

```
[ ] Yes
[x] No
```


apereverzin (approved):
It makes sense to move more code from BlobContainerClientProxy to BlobMover or to some new class.

PR #938, opened 2021-02-19
refactor-use blob mover upload

mdayican:
### Change description ###

refactor- use blob mover upload
later unused proxy methods will be deleted.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
```diff
@@ -39,4 +51,44 @@ public void dispatch(
...
+
+                    logger.info(
+                        "Streaming finished for  blob {} to Container: {}, Upload Duration: {} sec",
+                        sourceBlob.getBlobUrl(),
+                        destinationContainer,
+                        TimeUnit.NANOSECONDS.toSeconds(System.nanoTime() - startTime)
+                    );
+                    return;
+                }
+            }
```
apereverzin, line 82: Move these lines to `uploadEnvelope()` method and `InvalidZipArchiveException` from that method. This will make code better readable.

mdayican: where is this method?

apereverzin: It should be a new method.

mdayican: in that case that new method return the status so caller can throw InvalidZipArchiveException. I think it wouldn't make more readable

mdayican: I added new method but it  throws 'InvalidZipArchiveException'

apereverzin (approved):

src/main/java/uk/gov/hmcts/reform/blobrouter/services/storage/BlobDispatcher.java
apereverzin: For me - it is easier to understand.

PR #1021, opened 2021-04-13
refactor blob processing

mdayican:

https://tools.hmcts.net/jira/browse/DTSBPS-485

### Change description ###

try to reduce db call and simplify process

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
@@ -7,14 +7,12 @@ import com.azure.storage.blob.models.BlobItem;
 import org.slf4j.Logger;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.blobrouter.data.envelopes.Envelope;
-import uk.gov.hmcts.reform.blobrouter.data.envelopes.Status;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/tasks/processors/ContainerProcessor.java
```diff
@@ -76,68 +74,26 @@ private boolean isReady(BlobItem blobClient, String containerName) {
...
-    private Optional<Envelope> getLastEnvelope(BlobClient blobClient) {
+    private Optional<Envelope> findEnvelopeNotInCreatedStatus(BlobClient blobClient) {
         return envelopeService
-            .findLastEnvelope(blobClient.getBlobName(), blobClient.getContainerName());
+            .findEnvelopeNotInCreatedStatus(blobClient.getBlobName(), blobClient.getContainerName());
     }
 
-    private void leaseAndThen(BlobClient blobClient, Runnable action) {
+    private void leaseAndProcess(BlobClient blobClient) {
         leaseAcquirer.ifAcquiredOrElse(
```
apereverzin, line 94: After acquiring lease we need to ensure again that there is no envelope NOT in CREATE state - to eliminate any consequences of race condition.

mdayican: that part moved to blob processor.

apereverzin (approved):

PR #1104, opened 2021-06-16
StatisticsRepository - daily document counts

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-635.

In service layer results will be aggregated by weeks and months.

It is possible to aggregate by weeks using SQL (because number of seconds per week is the same), but number of seconds per month differs from month to month.

### Change description ###

```
[ ] Yes
[x] No
```


apereverzin:
Not needed

PR #1285, opened 2022-01-05
DTSBPS-827 - Added file_size

apereverzin:

### Change description ###

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V017__Add_file_size_envelopes.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE envelopes
+ADD COLUMN file_size INTEGER NOT NULL DEFAULT 0;
```
mdayican, line 2: do we need default 0 ?

apereverzin: I agree
Just realized - how will NULL be converted into int in Java code?

mdayican: object Integer should solve it.

mdayican (approved):

PR #1288, opened 2022-01-06
Added fileSize to Envelope + EnvelopeRepo

apereverzin:

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/blobrouter/data/envelopes/Envelope.java
+++ b/src/main/java/uk/gov/hmcts/reform/blobrouter/data/envelopes/Envelope.java
@@ -14,7 +14,7 @@ public class Envelope {
     public final Status status;
     public final boolean isDeleted;
     public final boolean pendingNotification;
-    public final int fileSize;
+    public final Long fileSize;
 
     public Envelope(
         UUID id,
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/hmcts/reform/blobrouter/data/envelopes/Envelope.java
```diff
@@ -24,7 +25,8 @@ public Envelope(
         Instant dispatchedAt,
         Status status,
         boolean isDeleted,
-        boolean pendingNotification
+        boolean pendingNotification,
+        Integer fileSize
```
mdayican, line 29: would Int enough ?

src/main/java/uk/gov/hmcts/reform/blobrouter/data/envelopes/Envelope.java
```diff
@@ -26,7 +26,7 @@ public Envelope(
         Status status,
         boolean isDeleted,
         boolean pendingNotification,
-        Integer fileSize
+        int fileSize
```
mdayican, line 29: I mean are you going to keep the size as Bytes?
If we process 4GB file int would not be enough.??

mdayican (approved):

PR #1291, opened 2022-01-07
DTSBPS-827 - EnvelopService updateFileSize

apereverzin:

### Change description ###

```
[ ] Yes
[x] No
```


mdayican (commented):
is it possible to write size while we are creating the record?

apereverzin:
I am afraid no because in BlobProcessor::getEnvelopeSupplier the envelope is created and BlobProcessor::handle zip file is dispatched, these methods are called in BlobProcessor::process.

Changing this logic is complicated and risky

apereverzin:
PR-1292

PR #1332, opened 2022-02-15
Configure privatelaw

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-839

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -100,6 +100,7 @@ service:
       sas-validity: 300 #In seconds
       target-storage-account: cft
       target-container: privatelaw
+      enabled: ${PRIVATELAW_ENABLED}
 
 queue:
   notifications:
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yaml
```diff
@@ -100,7 +100,6 @@ service:
       sas-validity: 300 #In seconds
       target-storage-account: cft
       target-container: privatelaw
-      enabled: false
```
mdayican, line 103: Pls make it env var and enable just for demo

src/main/resources/application.yaml
```diff
@@ -100,7 +100,7 @@ service:
...
+      enabled: ${PRIVATELAW_ENABLED}
```
mdayican, line 103: you need to add to values.yaml with false

mdayican (approved):

PR #2065, opened 2023-06-05
FACT-1378: cleanup suppressions

alivenichoppa:
https://tools.hmcts.net/jira/browse/FACT-1378

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/reform-scan-blob-router/values.preview.template.yaml
+++ b/charts/reform-scan-blob-router/values.preview.template.yaml
@@ -1,17 +1,20 @@
 java:
   secrets:
     STORAGE_ACCOUNT_NAME:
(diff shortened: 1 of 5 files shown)


charts/reform-scan-blob-router/values.preview.template.yaml
```diff
@@ -1,22 +1,22 @@
 java:
   secrets:
     STORAGE_ACCOUNT_NAME:
-      secretRef: storage-account-{{ .Release.Name }}-blobstorage
+      secretRef: storage-account-reform-scan
```
reespozzi, line 4: I think best to keep the {{ .Release.Name }} here and just lose the -blobstorage

amstevenson (approved):

PR #2322, opened 2025-03-21
FACT-2294 started dep updates

justiceia:
### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,4 +210,3 @@ You should get a response with status 200 and a sas token in the body.
 ## License
 
 This project is licensed under the MIT License - see the [LICENSE](LICENSE) file for details
-
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
   ext {
-    flywayVersion = '10.20.1'
+    flywayVersion = '10.22.0'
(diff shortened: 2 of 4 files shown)


RuthKirby (changes requested):
Can you double check the test dependencies? It seems like several can go up and blob router is behind orchestrator in AssertJ

justiceia:
> Can you double check the test dependencies? It seems like several can go up and blob router is behind orchestrator in AssertJ

bumping breaks it seems

RuthKirby (approved):

RuthKirby:
> > Can you double check the test dependencies? It seems like several can go up and blob router is behind orchestrator in AssertJ
> 
> bumping breaks smoke tests it seems

That's cool, if you can create a ticket for this and we will get to updating it as its own thing

justiceia:
> > > Can you double check the test dependencies? It seems like several can go up and blob router is behind orchestrator in AssertJ
> > 
> > 
> > bumping breaks smoke tests it seems
> 
> That's cool, if you can create a ticket for this and we will get to updating it as its own thing

created https://tools.hmcts.net/jira/browse/FACT-2362

amstevenson (approved):

